PR #38, opened 2026-04-19
feat(logger)!: unreserve the `version` field; user context flows through (closes #36)

### LINKED ISSUE #36: @centient/logger silently strips user-supplied `version` field from log context
## Summary

`@centient/logger` currently reserves the key `version` as a logger-instance meta-field and silently discards any user-supplied `version` in a log context. Two codepaths are involved:

1. **`Logger.ts:130-136`** — `buildEntry()` destructures `version: _v` out of the sanitized context and replaces it with `this.version` (the logger's own service version, defaulted to `"0.0.0"` if `createComponentLogger` wasn't given one).
2. **`format.ts:43-53`** — `formatPretty()` destructures `version: _v` out of the log entry before rendering the flat `key=value` tail, treating it as a "standard field" alongside `service`/`pid`/`hostname`.

Net effect: calling `logger.info({ version: "0.8.4", ... }, "...")` emits a line with no `version=` in it anywhere, and the JSON output carries the logger-instance version (often `"0.0.0"`) rather than what the caller passed.

## Observed impact

Surfaced in `@centient-labs/daemon@0.6.0` running inside `@centient-labs/maintainer@0.8.4`. Daemon startup banner:

```
02:29:21.267 [INFO] [core] Starting maintainer daemon name=maintainer port=8090 tls=false
03:20:42.995 [INFO] [core] maintainer daemon started name=maintainer port=8090 tls=false
```

The daemon passes `{ name: config.name, version: config.version, port: config.port, tls: !!config.tls }` to `logger.info(...)` — `version` is stripped. See the downstream tracking issue at `centient-labs/daemon#5` for production-incident context (missing version field made it hard to attribute log lines to specific maintainer releases during the v0.8.1 self-revert loop).

## Why the current behaviour is wrong

- The reserved `version` slot holds the *logger instance's* version, which is never actionable — nobody debugs by "which version of `@centient/logger` produced this line." Schema evolution, multi-version deployment correlation, and field serialization bugs are all better addressed by CHANGELOG/git tags or startup-once metadata, not every log entry.
- The *consumer's* service version (what a downstream dev actually wants to grep for in an incident) has no available field name, because the natural one is stealth-reserved.
- Pretty format intentionally hides some top-level fields (`service`, `pid`, `hostname`) to reduce clutter — that design is correct for those fields, but `version` is misclassified: it's a service identifier in practice, not a log-schema meta-field.

## Proposed fix

Preferred — **two related changes, shipped together:**

1. **Drop the reserved `version` slot from `LogEntry`.** The logger no longer emits its own version field; the `service` field already identifies the emitter. Remove `this.version`, remove the `options.version` constructor option (or deprecate it as a no-op for one release), remove the field from `types.ts::LogEntry`.
2. **Stop stripping `version` from user context** in both `Logger.ts::buildEntry` (remove the `version: _v` from the destructure) and `format.ts::formatPretty` (remove `version: _v` from the "standard fields" destructure).

This frees `version` for user context — the most intuitive name for "this service's semver at runtime" — and eliminates a field that was never earning its keep.

Alternative — **rename the reserved slot** (more conservative, still fixes user context):

1. Rename the logger-instance field from `version` to `schemaVersion` or drop the internal storage entirely.
2. Same formatter fix (stop stripping `version` from user context).

Either direction works; the drop-it variant is cleaner if the meta-field has no known consumer.

## Compatibility notes

- **Option-preferred (drop it):** The `version` constructor option on `createComponentLogger(...)` becomes unused. If any downstream consumer actually passes it today, they'd see it silently stop appearing in logs. A grep across `centient-labs/*` and `centient/*` packages before merge would catch callers; a deprecation pass-through for one minor (accept the option, emit a deprecation-once warning, ignore the value) would be safer.
- **Rename variant:** fully backwards-compatible on the user-facing API.
- **Log-format change is visible to downstream parsers** either way — if anyone greps logs for the current (useless) `version=0.0.0` style line, that lookup changes. Worth a `BREAKING CHANGE` or `feat!:` marker in the changeset.

## Reproduction

```ts
import { createComponentLogger } from "@centient/logger";

const logger = createComponentLogger("test", "component");
logger.info({ version: "1.2.3", name: "myapp" }, "starting");
// Expected: "... starting name=myapp version=1.2.3"
// Actual:   "... starting name=myapp"
```

## Downstream tracking

- `centient-labs/daemon#5` — the daemon-package workaround (planned: rename the daemon's own log fields from `version` to `appVersion` in two call sites). Per this issue's resolution, the daemon can revert its rename and go back to `version` once the logger is fixed — or keep `appVersion` and let consumers migrate on their own schedule.

Happy to open a PR for whichever direction you prefer.

owenjohnson:
Closes #36.

## Summary

\`@centient/logger\` was silently reserving \`version\` as a logger-instance meta-field and stripping any user-supplied \`version\` from log context — replacing it with \`this.version\` (defaulting to \`\"0.0.0\"\`). This PR removes the reservation. Callers who put \`version\` in context now see it flow through to the emitted entry like any other field.

Only \`service\`, \`component\`, and \`tool\` remain reserved. Everything else — including \`version\` — is caller-owned.

## Why this is safe as a clean removal

**Grep audit: zero callers.** I searched the entire \`centient-labs\` ecosystem for any consumer passing the \`version\` option to the logger factories. Result:

| Repo | Usages of \`createComponentLogger\`/\`createLogger\` | Passing \`version:\` | 
|---|---|---|
| \`centient-labs/centient-sdk\` | 5 internal (events, wal) + tests | 0 |
| \`centient-labs/daemon\` | 5 (cli, client, rate-limiter, web, core) | 0 |
| \`centient-labs/crucible\` | 6 (audit, benchmark, pipeline, discovery, roadmap, paginated-list) | 0 |
| \`centient-labs/centient\` | 5 (entry, Config, SessionBase, promotion-scorer, MetaKnowledgeEncryption) | 0 |
| \`centient-labs/soma\` | 1 (daemon logger singleton) | 0 |
| \`centient-labs/test-kit\` | 3 (mcp, http, containers) | 0 |

Every call is the default 2-arg form \`createComponentLogger(service, component)\`. Per the review discussion on #36, \"zero callers\" resolves to clean removal and a major bump rather than a deprecation shim.

### Removed
- \`LoggerOptions.version?: string\` (option no longer accepted)
- \`LogEntry.version: string\` (field no longer emitted)
- \`createTestLogger\`'s internal \`version: \"0.0.0-test\"\` default

### Behavior changes
- \`Logger.ts::buildEntry\` no longer strips \`version\` from sanitized context.
- \`format.ts::formatPretty\` no longer strips \`version\` from pretty-mode output — if you pass it in context, it appears in the tail like \`version=1.2.3\`.
- Child loggers no longer propagate the defunct \`this.version\`.

### Unchanged
- \`service\`, \`component\`, \`tool\` still reserved (they're genuinely logger-level identity and nobody complained about them).
- \`timestamp\`, \`level\`, \`message\`, \`pid\`, \`hostname\` still always-computed top-level fields.

## Migration

- Passing \`{ version: \"...\" }\` to any factory: **remove it.** The value wasn't appearing usefully in logs anyway. If you want per-line service version, put it in context: \`logger.info({ appVersion: \"1.2.3\" }, \"...\")\`.
- Reading \`entry.version\` on a captured \`LogEntry\`: the field is now \`undefined\` unless the caller explicitly put it in context. Use \`entry.service\` for emitter identity.
- Log-output parsers that expect a top-level JSON \`version\` field: adjust them. Nobody was parsing \`version=0.0.0\` usefully, but it's worth signaling.

## Scope

\`@centient/logger::AuditWriter\` has an equivalent \`this.version\` field on audit events (AuditWriter.ts:45, 57, 288). That's a **different concern**: audit events are a versioned schema payload, not a user-context field — callers don't pass \`version\` into audit events the same way they pass it into log context. I'm leaving AuditWriter alone here; if we want to re-examine it, that's a separate issue.

## Convention

The README now suggests \`appVersion\` as the centient-labs ecosystem convention (matches what \`@centient-labs/daemon\` already uses), but makes clear the logger doesn't enforce anything — pick whatever name is unambiguous to your log consumer.

## Verification

- Build: ✅ (workspace-wide)
- Lint: ✅ (tsc \`--noEmit\` across all packages)
- Tests: ✅ **481 pass, 14 skipped** (14 skipped are perf-gated per #32)
- Zero downstream compile breaks — the \`LogEntry\` type change is absorbed by the existing \`[key: string]: unknown\` index signature in consumer code.

## Post-merge

After this lands, the \`@centient-labs/daemon\` agent will bump its \`@centient/logger\` dep in a follow-up patch. The daemon is keeping \`appVersion\` as its field name (not reverting to \`version\`) — agreed in review: the constraint produced a more semantically-precise name, worth keeping.

### DIFF
--- a/.changeset/logger-unreserve-version.md
+++ b/.changeset/logger-unreserve-version.md
@@ -1,0 +1,45 @@
+---
+"@centient/logger": major
+---
+
+**BREAKING:** Unreserve the `version` field name. Removes the `version` option on `LoggerOptions` and the `version` slot on `LogEntry`. User-supplied `version` in log context now flows through to the emitted entry instead of being silently stripped. Closes #36.
+
+### What changed
+
+Previously, `@centient/logger` reserved `version` as a logger-instance meta-field. Two problems:
+
+1. **`Logger.ts::buildEntry`** destructured `version` out of user context and replaced it with `this.version` (defaulting to `"0.0.0"` when no option was supplied).
+2. **`format.ts::formatPretty`** stripped `version` from pretty-mode output before rendering the tail.
+
+Net effect: `logger.info({ version: "1.2.3" }, "...")` silently dropped the user's value and emitted nothing (pretty) or `"0.0.0"` (JSON). Observed on `@centient-labs/daemon@0.6.0` during the maintainer v0.8.1 incident (log lines with no attributable version).
+
+### The fix
+
+- Remove `LoggerOptions.version?: string`
+- Remove `LogEntry.version: string`
+- Remove the `version: _v` destructure in `Logger.ts::buildEntry`
+- Remove the `version: _v` destructure in `format.ts::formatPretty`
+- Remove `this.version` internal storage
+- `createTestLogger` no longer injects a `version: "0.0.0-test"` default
+
+Only `service`, `component`, and `tool` remain reserved. Everything else (including `version`) is yours.
+
+### Why major
+
+Zero callers in the centient-labs ecosystem passed the `version` option — verified via cross-org grep across `centient-sdk`, `daemon`, `crucible`, `centient`, `soma`, `test-kit`. Per the review discussion on #36, "zero callers" resolves to clean removal and a major bump rather than a deprecation shim.
+
+The log-output shape also changes: JSON entries no longer carry a top-level `version` field, and pretty-mode output no longer contains the useless `version=0.0.0` tail (it was being stripped, but any downstream parser that expected the JSON field is affected).
+
+### Migration
+
+- If you were passing `{ version: "..." }` to `createLogger` / `createComponentLogger` / etc.: remove it. The value wasn't useful in logs anyway. If you want per-line service version, pass it in context: `logger.info({ appVersion: "1.2.3", ... }, "...")`.
+- If you were relying on `entry.version` being populated in a log consumer: the field is gone. Use `service` (unchanged) to identify the emitter, or have the emitter put its version into context explicitly.
+- If your code reads `entry.version` on a captured log entry: the field is now `undefined` unless the caller put it in context.
+
+### Convention
+
+`@centient-labs/daemon` uses `appVersion` for the consuming app's semver. The centient-sdk README suggests following that convention unless your domain has a better-fitting name. See `packages/logger/README.md` §"Reserved top-level fields".
+
+### Scope
+
+This PR addresses the bug in `@centient/logger` only. `@centient/logger::AuditWriter` has a similar internal `this.version` on audit events — that's a separate schema-versioning concern for audit payloads, not a user-context field, so it's intentionally not touched here.
--- a/packages/logger/README.md
+++ b/packages/logger/README.md
@@ -39,6 +39,8 @@ The logger reserves these field names for its own top-level entry shape and stri
 
 > Prior to v1.0.0, `version` was silently reserved by the logger — any `version` in your context would be replaced with the logger-instance version (usually `"0.0.0"`). That reservation has been removed. See [issue #36](https://github.com/centient-labs/centient-sdk/issues/36).
 
+> **AuditWriter note:** `AuditWriter` (also exported from this package) emits its own top-level `version` field on emitted audit events. That value is the **audit-event schema version** maintained internally by the writer, not a user context field — distinct from the `Logger` reservation policy above. If you're reading raw audit records and see a `version` key, it's the audit schema, not anything the caller passed.
+
 ## API
 
 ### Logger
--- a/packages/logger/README.md
+++ b/packages/logger/README.md
@@ -19,10 +19,7 @@ pnpm add @centient/logger
 ```typescript
 import { createLogger, ConsoleTransport } from "@centient/logger";
 
-const logger = createLogger({
-  service: "my-service",
-  version: "1.0.0",
-});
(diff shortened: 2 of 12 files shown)


owenjohnson:
Pushed fix commit `b65e801` addressing all 6 findings from the 15:03 review — the review predates the commit. Summary:

- **[MED] api-contracts (#1):** real bug fix — `buildEntry` now strips `timestamp`, `level`, `message`, `pid`, `hostname` from user context alongside `component`/`tool`/`service`. Previously `...restContext` spread AFTER the computed fields, silently overriding them. New test confirms reserved fields in context cannot clobber computed values.
- **[MED] test-coverage (#2):** new test for `version` in `options.context` (baseContext path).
- **[MED] test-coverage (#3):** new test for `version` flow-through via `logger.child()`.
- **[LOW] test-coverage (#4):** new test pinning merge precedence — per-call `version` wins over baseContext `version`.
- **[LOW] api-contracts (#5):** dropped the unnecessary `as Partial<LogEntry>` cast.
- **[LOW] architectural (#6):** added README note that AuditWriter's internal `version` is the audit-schema version, separate from Logger's reservation policy.

485 tests pass (+ 4 new regression tests), 14 perf-gated skipped. Ready for re-review.

PR #51, opened 2026-04-21
fix(logger): flush buffered entries on FileTransport.close()

owenjohnson:
## Summary

`FileTransport.close()` was dropping any entries still in the in-memory buffer at close time. The ordering was:

```ts
this.closed = true;    // set first
this.flushSync();      // early-returns because closed is true
```

`flushSync()` guards against reentrancy with `if (... || this.closed) return;`, so the final drain in `close()` never wrote anything. Callers that wrote a handful of entries (fewer than `maxBufferSize`, so no size-triggered auto-flush) and then called `close()` without first calling `flush()` got an empty file.

The existing test suite masked this because every test pairs `await transport.flush(); await transport.close();`. I discovered it while working on PR #68 on [engram-server](https://github.com/centient-labs/engram-server/pull/68) — a bot review suggested using `await logger.close()` as a simple flush-before-assert pattern in a new test, and it unexpectedly failed. Tracing down exposed the ordering bug here.

## The fix

Swap the two lines in `close()` so `flushSync()` runs before `closed` is set. The `closed = true` guard still blocks new writes via `write()`'s own early-return, and the subsequent `writeStream.end(...)` still closes cleanly.

## Test

`close() flushes buffered entries without requiring a prior flush()` — writes one entry, calls `close()` directly (no `flush()`), asserts the entry is in the file. Deterministically fails on current main; passes with this fix.

## Verification

`make check` — tsc 0 errors, vitest **1174 passed / 14 skipped (1188 total)**. Logger package alone: **486 passed / 14 skipped**.

## Changeset

`.changeset/fix-file-transport-close-flush.md` — patch bump for `@centient/logger`.

### DIFF
--- a/.changeset/fix-file-transport-close-flush.md
+++ b/.changeset/fix-file-transport-close-flush.md
@@ -2,10 +2,14 @@
 "@centient/logger": patch
 ---
 
-Fix `FileTransport.close()` dropping buffered entries.
+Fix `FileTransport.close()` dropping buffered entries and harden the close path against in-flight failures.
 
-`close()` previously set `this.closed = true` before calling `flushSync()`. Because `flushSync()` early-returns when `this.closed` is true, any entry still in the in-memory buffer at the time of close was silently discarded. Callers that wrote a handful of entries (fewer than `maxBufferSize`, so no size-triggered auto-flush) and then called `close()` without an intervening `flush()` would get an empty file.
+The original bug: `close()` set `this.closed = true` before calling `flushSync()`. Because `flushSync()` early-returns when `this.closed` is true, any entry still in the in-memory buffer at close time was silently discarded. Masked in practice by call sites pairing `await transport.flush(); await transport.close();`. The fix flushes before marking closed, and adds a regression test that calls `close()` without a prior `flush()`.
 
-In practice this was masked by every call site pairing `await transport.flush(); await transport.close();` — including the test suite. Surfacing the bug therefore required a test that calls `close()` on its own; adding one demonstrated the regression and this PR fixes it.
+Additional hardening for the close path:
+- Wait for in-flight rotation in a try/catch so a rotation rejection cannot leave the transport in a half-open zombie state with `this.closing` pinned to a permanently-rejected promise.
+- Reset `this.closing = null` after `_doClose()` settles so a failed close attempt does not pin a rejected promise on the instance.
+- Add a reject path to the `writeStream.end()` Promise so a stream error during finalization cannot leave the close awaiter pending forever.
+- Preserve the buffer on a throwing `writeStream.write()` (write before clearing) so entries are not silently dropped.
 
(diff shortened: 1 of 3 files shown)


owenjohnson:
Pushed `0670748` addressing all four findings.

**[MEDIUM] error-handling — flushSync() may throw and leave the transport half-open.** Fixed. Wrapped the flush+mark-closed sequence in try/finally so `this.closed = true`, `clearInterval(flushTimer)`, and `this.flushTimer = null` all run even if `flushSync()` throws (e.g. ERR_STREAM_DESTROYED). The half-open pathology you described is now closed — the timer stops and new writes hit the `closed` guard even on the error path.

**[LOW] test-coverage — toContain is too loose.** Fixed. The test now splits the file into lines, asserts exactly one line, and parses the JSONL to validate both structure and the expected `message` field. A malformed-JSONL partial write now fails loudly.

**[LOW] architectural-consistency — multi-line comment above flush.** Fixed. Collapsed to a single line: `// Must flush before marking closed: flushSync() early-returns when this.closed is true.`

**[LOW] test-coverage — in-body block comment in the test.** Fixed. Removed; the test description plus the changeset carry the history.

`make check` on `0670748`: tsc 0 errors, vitest 1174 passed / 14 skipped.

owenjohnson:
Pushed `3f67873` addressing all six findings.

**[HIGH] error-handling — stream-close skipped when flushSync() throws.** Fixed. `close()` now captures the flush error in a local, runs the `writeStream.end(...)` path unconditionally, and re-throws at the very end. File handle is no longer leaked on an `ERR_STREAM_DESTROYED` throw.

**[HIGH] error-handling — flushSync drops buffer before write.** Fixed. Swapped the order: the joined string is written first, then `this.buffer = []` only runs on success. A throwing `writeStream.write()` now leaves the buffer intact for a later flush / retry instead of silently dropping entries.

**[MEDIUM] error-handling — pendingFlush rejection bypassed close.** Fixed. `await this.pendingFlush` is now inside a try/catch; the cleanup that follows runs regardless of whether the in-flight flush errored.

**[MEDIUM] test-coverage — no test for the flushSync-throw cleanup.** Fixed. New test `close() completes cleanup even when flushSync() throws` stubs `writeStream.write` to throw deterministically, asserts `close()` rejects, AND that `closed === true` and `flushTimer === null` after. Guards the resource-leak fix above.

**[MEDIUM] test-coverage — write-after-close not exercised.** Fixed. New test `write() after close() is a no-op` writes, closes, writes again, calls flush, asserts the post-close entry did not reach the file. Pins the `write()` early-return contract.

**[LOW] api-contracts — concurrent close protection.** Fixed. Added a `closing: Promise<void> | null` latch. Concurrent callers now share the same in-flight close promise, matching the `rotating` / `pendingFlush` pattern already in the file. New test `close() is idempotent under concurrent calls` fires three parallel `close()` calls and asserts single on-disk entry (no double-write, no double-close).

`make check` on `3f67873`: tsc 0 errors, vitest 1177 passed / 14 skipped (+3 new cases, all green).

PR #56, opened 2026-04-25
feat(sdk): add crystals.related(id)

owenjohnson:
## Summary

- Adds `client.crystals.related(id)` to the SDK, wrapping `GET /v1/crystals/:id/related`. Returns the paginated edge envelope unwrapped to `{ edges, total, hasMore }` — mirrors the typing pattern of `crystals.list()`.
- Adds a changeset for a `@centient/sdk` patch bump (1.7.0 → 1.7.1).
- Adds two unit tests: happy-path envelope unwrap and id URL-encoding.

## Caveat to flag (not fixing here)

The route is documented as "Get semantically related nodes" but the current server implementation calls `getNodeRelationships()` and returns graph edges (incoming + outgoing), not embedding-similar nodes. The SDK method is named `related()` (matches the URL) and the JSDoc spells out the actual behaviour.

UI consumers (engram-web search-polish PR is the first one) should label the affordance "Related items" / "Connections" rather than "Similar" until the server-side semantics are settled.

## Test plan

- [x] `pnpm test` — 393/393 pass (2 new)
- [x] `pnpm run build` — clean
- [x] `pnpm run lint` — clean

### DIFF
--- a/.changeset/crystals-related.md
+++ b/.changeset/crystals-related.md
@@ -1,0 +1,7 @@
+---
+"@centient/sdk": patch
+---
+
+Add `crystals.related(id)` method that wraps `GET /v1/crystals/:id/related` and returns the paginated edge envelope as `{ edges, total, hasMore }`.
+
+The current server implementation returns graph neighbours (incoming + outgoing edges), not embedding-similarity matches — callers should label UI accordingly. Mirrors the typing pattern of `crystals.list()`.
--- a/packages/sdk/src/resources/crystals.ts
+++ b/packages/sdk/src/resources/crystals.ts
@@ -436,6 +436,29 @@ export class CrystalsResource extends BaseResource {
   }
 
   /**
+   * Get edges connected to a crystal node (graph neighbours).
+   *
+   * Returns all relationships where the node is either the source or target,
+   * flattened into a single list. The current implementation returns graph
+   * edges (incoming + outgoing), not embedding-similar nodes.
+   */
+  async related(id: string): Promise<{
+    edges: KnowledgeCrystalEdge[];
+    total: number;
+    hasMore: boolean;
+  }> {
+    const response = await this.request<ApiSuccessResponse<KnowledgeCrystalEdge[]>>(
+      "GET",
+      `/v1/crystals/${encodeURIComponent(id)}/related`
+    );
+    return {
+      edges: response.data,
+      total: response.meta?.pagination?.total ?? response.data.length,
+      hasMore: response.meta?.pagination?.hasMore ?? false,
+    };
+  }
+
+  /**
    * List knowledge crystal nodes with optional filters.
    *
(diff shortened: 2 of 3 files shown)


owenjohnson:
## Triage of maintainer findings

Pushed [d6a91ce](https://github.com/centient-labs/centient-sdk/commit/d6a91ce) addressing #1 (HIGH) and #4 (MEDIUM). Other findings intentionally skipped — explanations below — and I'm flagging this for **manual merge** rather than chasing further rounds.

### Applied
- **#1 [HIGH] fallback branches untested** → new test asserts `total = data.length` and `hasMore = false` when `meta.pagination` is absent. Exercises both `??` branches.
- **#4 [MEDIUM] hasMore: true propagation** → new test asserts `meta.pagination.hasMore = true` round-trips to `result.hasMore = true`.

### Skipped — with rationale

- **#2 [MEDIUM] pagination params on `related()`.** The server route at `apps/engram/src/server/routes/crystals.ts:2742-2773` does not paginate — it returns the full flat list `[...outgoing, ...incoming]` and sets `limit = allRelated.length`, so the envelope can never produce `hasMore: true` in practice. Adding `offset`/`limit` params to the SDK would be misleading. The handoff for the consuming engram-web PR explicitly says: *"Don't add params yet — the route doesn't accept any beyond id."* The `hasMore` field is kept to faithfully mirror the envelope shape (and the new test in #4 covers the future case where the server starts paginating).
- **#3 [MEDIUM] silent fallback to `data.length`.** This is the same pattern used by every other unwrapped paginated method in this file: `crystals.list()` (L492), `versions().list()` (L186), `items().list()` (L107), `hierarchy().getChildren()` (L283), etc. Changing it for `related()` alone would create inconsistency; cross-cutting cleanup belongs in a separate PR.
- **#5 [MEDIUM] error-response tests.** Error propagation is `BaseResource.request()`'s responsibility and is tested at that layer. No other resource method (`get`, `list`, `versions().list()`, etc.) has per-method error tests — adding one here only would suggest `related()` is special, which it isn't.
- **#6 [LOW] `createMockEdge` missing fields.** The helper is pre-existing and shared with all other crystal-edge tests. Updating it touches unrelated tests; out of scope for this PR. Worth a follow-up but doesn't block.
- **#7 [LOW] empty-id guard.** None of the existing id-based methods (`get`, `delete`, `versions().get`, etc.) have one. Cross-cutting concern; if it's adopted, it should land at the `BaseResource` level for all id-bearing routes at once.

### Status
- `pnpm test`: 395/395 (393 existing + 2 new)
- `pnpm run build`: clean
- `pnpm run lint`: clean

Per "don't chase the spiral" guidance from prior maintainer-bot rounds, requesting manual merge here. The two real correctness gaps (#1 and #4) are now covered; the rest are stylistic / cross-cutting and would expand scope beyond the tier-4 #73 item this is unblocking.

PR #58, opened 2026-06-04
feat(sdk)!: align with engram-server 0.34.0

owenjohnson:
## Summary

Brings `@centient/sdk` into alignment with **engram-server 0.34.0**. A full review of `apps/engram/CHANGELOG.md` surfaced the SDK-relevant deltas in 0.33.0 and 0.34.0. The headline: the SDK **sync resource was last aligned at server v0.22.4** and predated the 0.34.0 `@hono/zod-openapi` envelope migration, so several methods were mis-modeled against the real wire contract (verified against the live server routes + `sync-schemas.ts`).

## Changes

**Sync resource — realigned to the real contract** (`src/resources/sync.ts`)
- `SyncStatus` → `{ instanceId, schemaVersion, peersCount, activeLinksCount, changelogSize }`
- `push`/`pushTo` → `counts` is per-entity `{ inserted, updated, skipped }`; dropped phantom `success`
- `push` sends **NDJSON** (`application/x-ndjson`); `SyncChange` matches the serialized changelog-entry shape
- `pull` parses the **NDJSON stream** (the one route never enveloped); `sinceSeq` now required
- `pullFrom` returns apply counts `{ entriesStreamed, maxSeq, duration }`, not a change array
- `listConflicts` reads the nested `{ conflicts, total }` (no `hasMore`)

**Added**
- `skipEmbedding` on `crystals.create()` (engram-server#763)
- `maintenance.vacuum({ full? })` (engram-server#766) — bare non-enveloped response, admin-gated `full`
- `EngramClient._requestRawBody()` helper for the NDJSON push

**Documented (no API change)**
- audit `ingest`/`ingestBatch` 202 responses (0.33.0)
- 409 `SYNC_SCHEMA_VERSION_MISMATCH` details on `EngramError.details`
- `MIN_SERVER_VERSION` stays `0.31.0` with per-feature 0.34.0 notes

## Breaking changes

Sync resource public types changed (`SyncStatus`, `SyncCounts`, `SyncPushResult`/`SyncPullResult`, `SyncChange`); `pull()` requires `sinceSeq`; `pullFrom()` returns counts; `listConflicts()` drops `hasMore`. Captured in a **major** changeset.

## Verification
- `npm run build` ✅
- `npm test` ✅ — **400 passed**
- `npm run lint` ✅ (`tsc --noEmit`)


owenjohnson:
Verdict: **changes-requested** before publishing this SDK major.

**HIGH** `packages/sdk/src/resources/maintenance.ts:91` and `packages/sdk/src/resources/maintenance.ts:103`  
`tombstoneCleanup()` and `changelogCompact()` still unwrap `response.data`, but the engram-server 0.34 maintenance routes return bare JSON objects for these endpoints, the same way the new `vacuum()` method correctly handles its bare response. Against a 0.34 server these two methods return `undefined`, so the incident path this SDK release is meant to support remains broken.

Recommended fix: model all three maintenance endpoints as bare responses:
- `return this.request<TombstoneCleanupResult>("POST", "/v1/maintenance/tombstone-cleanup", params)`
- `return this.request<ChangelogCompactResult>("POST", "/v1/maintenance/changelog-compact", params)`
- keep `vacuum()` bare as implemented

Update the maintenance tests to mock `{ deleted, warnings, dryRun }` and `{ deleted, belowSeq, dryRun }` directly, not `{ data: ... }`.

**HIGH** `packages/sdk/src/resources/sync.ts:76` and `packages/sdk/src/resources/sync.ts:150`  
The sync peer subresource is still modeled against `{data: ...}` envelopes and an optional `apiKey`, but the 0.34 server peer routes use bespoke bare envelopes: create/get/link/pause/resume return `{ peer }`, list returns `{ peers }`, delete returns `{ removed: true, name }`, and create requires `apiKey`. As written, `sync.peers.create/list/get/delete()` return `undefined` or send an invalid create body, while tests still mock the old `{ data: ... }` shape.

Recommended fix: change `CreatePeerParams.apiKey` to required and unwrap the actual peer envelopes:
- create/get/link/unlink/pause/resume: request `{ peer: SyncPeer }`
- list: request `{ peers: SyncPeer[] }`
- delete: return `{ removed: true; name: string }` or keep `Promise<void>` intentionally, but do not expose `{ deleted: true }`

Update `sync.test.ts` to use bare peer responses and include `apiKey` in create params.

**MEDIUM** `packages/sdk/src/resources/sync.ts:93` and `packages/sdk/tests/resources/sync.test.ts:44`  
`SyncCounts` and `SyncChange.entityType` are left as unconstrained strings, and the new tests use `"crystals"`. The server 0.34 sync entity enum is `"knowledge_crystals" | "knowledge_crystal_edges" | "sessions" | "session_notes"`; sending `"crystals"` in NDJSON will not match the server table/allowlist contract. This lets callers compile requests that fail at runtime, and the tests now document an invalid entity type.

Recommended fix: add and export a `SyncEntityType` union matching the server enum, use it for `SyncChange.entityType`, `SyncPullParams.entityTypes`, and `SyncCounts`, and update tests/examples to use `knowledge_crystals`.

**MEDIUM** `packages/sdk/src/resources/index.ts:206` and `packages/sdk/src/resources/index.ts:230`  
New public types are not exported from the package root. `SyncCounts`, `SyncPullResult`, `VacuumParams`, and `VacuumResult` are declared in resource modules but omitted from `resources/index.ts`, so consumers cannot import them from `@centient/sdk` despite the new public methods returning/accepting them.

Recommended fix: export these types from `packages/sdk/src/resources/index.ts` and re-export them from `packages/sdk/src/index.ts` alongside the existing sync and maintenance types.

I did not rerun the suite; this review is based on PR #58 source plus the engram-server 0.34 route/schema contracts.

owenjohnson:
Addressed all 8 findings in 31c45f2.

| # | Sev | Finding | Resolution |
|---|-----|---------|------------|
| 1 | HIGH | `pull()` NDJSON `JSON.parse` had no try/catch | Wrapped per-line; re-thrown as `NetworkError` with line index + 200-char excerpt |
| 2 | MED | `push()` missing trailing newline | Non-empty payloads now newline-terminate every record (`…join("\n") + "\n"`); empty → `""`. Test body assertion updated |
| 3 | MED | No malformed-NDJSON test | Added: valid line + truncated line → asserts `NetworkError` |
| 4 | MED | No 409 schema-mismatch propagation test | Added: asserts `EngramError` with `code: SYNC_SCHEMA_VERSION_MISMATCH`, `statusCode: 409`, and `details: { peerVersion, ourVersion }` |
| 5 | MED | `vacuum()` bare-object assumption fragile | Now unwraps `.data` if a future server envelopes it, then validates `vacuumed` is an array and throws `EngramError` on shape drift; comment refs engram-server#766 |
| 6 | LOW | No vacuum 403 test | Added: `vacuum({ full: true })` → asserts `EngramError` on `AUTH_FORBIDDEN` |
| 7 | LOW | `_requestRawBody` retried `SyntaxError` from `json()` | Fast-fail on `SyntaxError` with a descriptive `NetworkError` (no retry budget burn) |
| 8 | LOW | `_requestRawBody` `@internal` not enforced | Strengthened JSDoc: "NOT part of the public API. No semver guarantees; may change or be removed in any release." |

Verification: `npm run build` ✅ · `npm test` ✅ **403 passed** (+3) · `npm run lint` ✅

owenjohnson:
Verdict: **changes-requested**.

Follow-up after the `31c45f2` fixes and the primary coding agent reply: the maintainer-bot findings appear to have been addressed, but the independent API-contract findings are still unresolved. The PR still should not be published as the 0.34 SDK alignment.

**HIGH** `packages/sdk/src/resources/sync.ts:280`, `322`, `340`, `361`, `386`, `402`, plus peers at `151-230`  
The SDK still models most sync responses as `{ data }`, but `engram-server/main` explicitly preserves bare sync shapes. This means `sync.push()`, `getStatus()`, `pushTo()`, `pullFrom()`, `listConflicts()`, `resolveConflict()`, and `sync.peers.*` return `undefined` or malformed values against the actual server. The PR changeset claim that `/v1/sync` migrated to standard `{success,data}` is false.

Recommended fix: model the real bare shapes:
- `push`: `{ success, counts, conflicts, duration }`
- `status`: bare status object
- `push-to` / `pull-from`: bare service result
- `conflicts`: `{ conflicts, total }`
- `resolve`: `{ success, conflict }`
- peers: `{ peer }`, `{ peers }`, `{ removed: true, name }`

Update the sync tests so they mock the server contract instead of `{ data: ... }` envelopes.

**HIGH** `packages/sdk/src/resources/maintenance.ts:92` and `packages/sdk/src/resources/maintenance.ts:104`  
`tombstoneCleanup()` and `changelogCompact()` still unwrap `response.data`, but server 0.34 returns bare maintenance objects. The new vacuum method was hardened, but the existing incident-related maintenance methods remain broken.

Recommended fix: request/return `TombstoneCleanupResult` and `ChangelogCompactResult` directly, and update tests currently mocking `{ data: ... }`.

**MEDIUM** `packages/sdk/src/resources/sync.ts:77`, `91`, `98`, `126`; `packages/sdk/tests/resources/sync.test.ts:45`, `58`, `105`, `149`, `236`  
Sync entity types remain unconstrained strings, and tests still use invalid `"crystals"`. Server enum is `knowledge_crystals | knowledge_crystal_edges | sessions | session_notes`.

Recommended fix: add/export `SyncEntityType`, use it for `SyncChange.entityType`, `SyncPullParams.entityTypes`, `SyncCounts`, and tests.

**MEDIUM** `packages/sdk/src/resources/index.ts:206` and `230`; `packages/sdk/src/index.ts:212` and `230`  
New public types are still not exported from the package root: `SyncCounts`, `SyncPullResult`, `VacuumParams`, `VacuumResult`.

Recommended fix: export them from `resources/index.ts` and `src/index.ts`.

**LOW** `packages/sdk/src/client.ts:529`  
`_requestRawBody()` parses JSON before checking `response.ok`. The SyntaxError fast-fail fix means a non-JSON 500 from `/sync/push` becomes a non-retried `NetworkError`, bypassing the 5xx retry/error parsing path.

Recommended fix: check `response.ok` first, retry 5xx before body parsing where possible, and use JSON/text fallback for error bodies.

owenjohnson:
Round 2 in 0f5c2e1. I verified every response shape against `engram-server` `origin/main` source before acting — the follow-up was **partly right and partly wrong**, so here's the evidence per finding.

### Accepted (fixed)

**Peers are bare — RIGHT.** `links.ts` on `origin/main` has **zero `success()` calls**; handlers return `{ peer: toPublicPeer(...) }`, `{ peers: [...] }`, and `{ removed: true, name }`. Fixed `SyncPeersResource`: `create`/`get` read `.peer`, `list` reads `.peers`, `delete` now returns `{ removed, name }` (was `{ deleted: true }`). Peer tests updated.

**Maintenance is bare — RIGHT.** `tombstoneCleanup`/`changelogCompact` handlers return bare `{ deleted, ... }` (not `success()`), so the SDK's `.data` unwrap was yielding `undefined` against a real server. Both now read the bare result directly (`vacuum` was already correct). Tests updated to bare mocks.

**`SyncEntityType` enum — RIGHT.** Added + exported `SyncEntityType = knowledge_crystals | knowledge_crystal_edges | sessions | session_notes`; wired through `SyncChange.entityType`, `SyncPullParams.entityTypes`, `SyncCounts`. Fixtures now use valid values (was `"crystals"`).

**Root exports — RIGHT.** Exported `SyncEntityType`, `SyncCounts`, `SyncPullResult`, `VacuumParams`, `VacuumResult` from `src/index.ts` + `src/resources/index.ts`.

**`_requestRawBody` ordering — RIGHT.** Now checks `response.ok` **before** parsing: a non-JSON 5xx body goes through the retry/error path (text fallback); only a non-JSON 2xx fast-fails.

### Rejected (with evidence)

**"sync.ts proper returns bare shapes / the changeset claim is false" — INCORRECT.** On `origin/main`:
- `apps/engram/src/types/api-response.ts` → `success<T>(data) => { success: true, data }`
- `apps/engram/src/server/routes/sync.ts` calls `success(...)` **7 times** — `status`, `conflicts` (`success({conflicts,total})`), `resolve`, `push`, `push-to`, `pull-from` all return `{ success, data }`.
- The 0.34.0 CHANGELOG documents this migration with an explicit before/after diff for `GET /v1/sync/status`.

So the SDK's `.data` unwrap for `getStatus`/`listConflicts`/`push`/`pushTo`/`pullFrom`/`resolveConflict` is correct, and the changeset claim is accurate. The finding conflated the **enveloped** `sync.ts` routes (cited lines 280/322/340/361/386/402) with the **bare** peers subtree in `links.ts` — only the latter (and maintenance) are bare, and those are now fixed. Changeset updated to spell out the split.

Verification: `npm run build` ✅ · `npm test` ✅ **403 passed** · `npm run lint` ✅

owenjohnson:
Round 3 in 8b07585. Addressed the 7 refinements (0 CRIT / 0 HIGH; 3 MED / 4 LOW) — 5 applied, 2 LOWs with rationale.

### Applied
- **[MED] vacuum dual-mode** → removed the speculative `?? (raw as VacuumResult)` unwrap; now `request<VacuumResult>` + the shape guard, consistent with `tombstoneCleanup`/`changelogCompact` (all three maintenance routes are bare). Also dropped the now-dead `ApiSuccessResponse` interface in `maintenance.ts`.
- **[MED] peers link/unlink/pause/resume tests** → mock the bare `{ peer }` shape instead of `{ data: undefined }`.
- **[MED] vacuum guard test** → added a case mocking `{ full: false }` (no `vacuumed`) → asserts `EngramError`.
- **[LOW] empty-push fixture** → fully conforming `counts` with all four `SyncEntityType` keys.
- **[LOW] pull body coverage** → existing test now asserts `body: {"sinceSeq":null}`; added a second case asserting `sinceSeq:"42"` + `entityTypes` are serialized.

### Rationale (not changed)
- **[LOW] `_requestRawBody`/`_requestRaw` `public` + `@internal`** — these must stay `public`: resource classes call `this.client._requestRaw(...)` across instances, which `private`/`protected` would forbid. The `@internal` JSDoc (now explicit about "no semver guarantees") is the intended signal; tightening the surface needs a separate internal-module refactor, out of scope for this PR.
- **[LOW] `_requestRawBody` error-parsing vs `_requestRaw`** — the `text()` → `JSON.parse()` order in `_requestRawBody` is deliberate: it reads the body exactly once. `_requestRaw`'s `json()` → `text()`-fallback pattern can hit "body already read" when `json()` partially consumes a non-JSON stream. As the finding itself notes, both handle the non-JSON-content-type case correctly, so this is a consistency nit; I'd rather not port the riskier double-read pattern. Happy to align `_requestRaw` to the single-read approach in a follow-up if preferred.

`npm run build` ✅ · `npm test` ✅ **405 passed** · `npm run lint` ✅

---

**⚠️ Operator note (infra, not code):** the latest maintainer review **escalated** — Phase 1 failed because `make check` **timed out at 8 min** and a **flaky `@centient/logger` test** (`FileTransport > buffered writing > should flush when buffer reaches maxBufferSize`, expected 3 got 1) failed. Both are **unrelated to this PR** (I didn't touch `@centient/logger`, and the SDK package's own suite is green). The bot flagged it as a bot-host issue with no author-actionable fix. This will likely keep blocking a clean Phase-1 pass until the host timeout and/or the logger flake are addressed.

owenjohnson:
Pushed 23f3a52 to de-flake the `@centient/logger` `FileTransport` test that contributed to the Phase-1 gate failure.

**Root cause:** `should flush when buffer reaches maxBufferSize` raced a fixed **50ms** sleep against the fire-and-forget async flush the 3rd `write()` schedules (rotation check + stream write). Under load that flush exceeds 50ms, so `readFileSync` intermittently saw an empty file — exactly the bot's failure (`expected [ '' ] to have a length of 3 but got 1`).

**Fix (test-only):** replaced the fixed sleep with a bounded poll (~2s) for the entries to land. The test still does **not** call `flush()` — it verifies the auto-flush-on-`maxBufferSize` path fires on its own. Stable across repeated runs; full logger suite green (490 passed), build + lint clean. No changeset (no public-API change).

This is cross-package (in `@centient/logger`, not the SDK) but folded into this PR because that flake was blocking this PR's review gate. The other half of the escalation — `make check` timing out at 8 min — is host-side and still needs operator attention.

owenjohnson:
Round 4 in 0e40e1e. **`make check` passed last round** (1192 tests) — the de-flake worked and the timeout didn't recur, so this is a pure code/test-quality pass. 7 of 8 applied; 1 LOW with rationale.

### Applied
- **[HIGH] vacuum `full` guard** → the shape guard now also requires `typeof result.full === "boolean"`, so `{ vacuumed: [] }` no longer returns `full: undefined`.
- **[HIGH] 409 mock self-consistency** → the `SYNC_SCHEMA_VERSION_MISMATCH` test now serializes the *same* error body through both `json()` and `text()`, so it holds under either error-parse strategy.
- **[MED] push/pushTo counts guard** → both validate all four `SyncEntityType` keys are present and throw `EngramError` on drift, matching the vacuum precedent.
- **[MED] `_requestRawBody` error-path tests** → added via `push()`: 5xx-then-retry-success, retry-budget exhaustion, `AbortError`→`TimeoutError`, fetch-failure→`NetworkError`.
- **[MED] pull empty-stream coverage** → `pull()` returns `[]` for `"\n"`, `"  \n  "`, and multi-blank-line bodies.
- **[LOW] count fixtures** → push/pushTo mocks now carry all four entity-type keys.
- **[LOW] logger poll** → re-throws non-`ENOENT` read errors (no longer masks `EACCES`/`EIO`) and widens the budget to ~5s.

### Rationale (not changed)
- **[LOW] `_requestRawBody`/`_requestRaw` `public` in `.d.ts`** (3rd raise) — these **must** be `public`: resource classes call `this.client._requestRaw(...)` on a *held reference*, not via inheritance, so `protected`/`private` would forbid it. The suggested `src/internal/http.ts` extraction is a real, reasonable cleanup, but it means threading all client config (`baseUrl`/`apiKey`/`timeout`/`retries`/`retryDelay`/`sleep`) through module-level functions and touching the core request path used by **every** resource — disproportionate and risky to land inside this alignment PR for a cosmetic `.d.ts` concern. I've filed it as a follow-up rather than churn the hot path here. The `@internal` JSDoc (now explicit about no semver guarantees) is the intended signal in the meantime.

`npm run build` ✅ · SDK `npm test` ✅ **410 passed** · logger transports stable 3/3 · `npm run lint` ✅ (both packages)

owenjohnson:
The `_requestRaw`/`_requestRawBody` internal-module extraction (the one outstanding LOW) is tracked in #60.

owenjohnson:
Round 5 in e3718fe. The bot dropped the prior #60 LOW (rationale accepted) and raised 9 new findings — mostly generalizing the round-4 guard pattern. 8 applied; 1 HIGH rejected with evidence.

### Applied
- **[HIGH] explicit no-retry** → `_requestRawBody` now short-circuits a self-raised `NetworkError` in the catch, so the non-JSON-2xx no-retry contract holds even if the error hierarchy changes.
- **[MED] `assertFullCounts` numerics** → each entity-type entry must have numeric `inserted`/`updated`/`skipped`.
- **[MED] maintenance guards** → `tombstoneCleanup`/`changelogCompact` now guard `deleted` is a number (parity with `vacuum`), so an enveloped/regressed body fails loudly.
- **[MED] pull error-path tests** → added `_requestRaw error handling (via sync.pull)`: 5xx retry, `AbortError`→`TimeoutError`, fetch-failure→`NetworkError`.
- **[MED] empty-push round-trip** → the empty-push test now asserts `result.conflicts` and `result.counts.*` (guard passes on a zero-entry response).
- **[LOW] pull line validation** → each parsed `SyncChange` must carry `seq`/`entityType`/`entityId`/`operation` → `NetworkError` on drift.
- **[LOW] maintenance negative tests** → `tombstoneCleanup`/`changelogCompact` reject an enveloped `{ data }` body.
- **[LOW] peers discard** → documented the intentional `{ peer }` discard on link/unlink/pause/resume.

### Rejected (with evidence)
- **[HIGH] "_requestRawBody doesn't retry 429 like request()"** — verified false. On `main`, **all three** request paths gate retry on `response.status >= 500` only (`request()` ~L619, `_requestRaw` ~L450, `_requestRawBody` ~L540); **none** retry on 429. So `_requestRawBody` already mirrors `request()` — there is no divergence. Adding 429 retry (ideally honoring `Retry-After`) is a reasonable *client-wide* enhancement, but it's a separate design decision, not a bug in this method, and blindly retrying 429 without backoff could hammer a rate-limited server. Happy to file it as a follow-up if you'd like it across the client.

`npm run build` ✅ · SDK `npm test` ✅ **415 passed** · `npm run lint` ✅

---

A note on convergence: the substantive contract (envelopes, bare shapes, NDJSON, skipEmbedding, vacuum) has been settled since round 2 (0 CRIT/0 HIGH there). Rounds 4–5 have been defensive-depth guards + test parity, which I've applied where cheap and correct. If subsequent rounds keep expanding runtime validation to every field/method, I'd suggest drawing the line and tracking further hardening as a follow-up rather than growing this alignment PR indefinitely — open to your call on that.

owenjohnson:
Round 6 in ada9b86 — all 6 findings applied (0 CRIT / 0 HIGH; 4 MED / 2 LOW).

- **[MED] push/pushTo null-safe data** → `requireData()` narrows the enveloped `data` before reading `.counts`, so `{ data: null }` throws a structured `EngramError`, not a `TypeError`.
- **[MED] listConflicts guard** → asserts `conflicts` is an array and `total` is a number (rejects an old `{ data: [] }` / null-data regression).
- **[MED] peers guards** → `requirePeer()` for `create()`/`get()` (`{ peer }`) and an array check for `list()` (`{ peers }`).
- **[MED] non-JSON 2xx test** → a 200 OK with a non-JSON body throws `NetworkError` and is **not** retried (asserts `fetch` called exactly once under `retries: 3`).
- **[LOW] pull createdAt** → added to the per-line required-field guard.
- **[LOW] pull entityType membership** → validated against `SYNC_ENTITY_TYPES`.

`npm run build` ✅ · `npm test` ✅ **416 passed** · `npm run lint` ✅

---

**Capping further hardening here.** The substantive 0.34.0 contract (envelopes, bare shapes, NDJSON, `skipEmbedding`, `vacuum`) has been settled since round 2 — rounds 4–6 have been defensive-depth guards, now applied to every method this PR touches. Extending exhaustive runtime shape-validation to the remaining untouched resources (`getStatus`, `resolveConflict`, `pullFrom`, and other resources) is tracked as a follow-up in **#62** rather than continuing to grow this alignment PR. I'd ask that any further round of the same pattern be deferred to #62 so #58 can land on the contract work it's actually about.

PR #61, opened 2026-06-05
test(logger): use vi.waitFor for FileTransport maxBufferSize auto-flush poll

owenjohnson:
## Summary

De-flakes `@centient/logger`'s `FileTransport > buffered writing > should flush when buffer reaches maxBufferSize`.

**Root cause:** the test raced a fixed **50ms** `setTimeout` against the fire-and-forget async flush that the 3rd `write()` schedules (rotation check + stream write). Under load that flush exceeds 50ms, so `readFileSync` intermittently saw an empty file — `expected [ '' ] to have a length of 3 but got 1`. This surfaced in the maintainer bot's Phase-1 `make check`.

**Fix (test-only):** replace the fixed sleep with a bounded poll (up to ~5s) for the entries to land. The test still does **not** call `flush()` — it verifies the auto-flush-on-`maxBufferSize` path fires on its own. The poll re-throws any non-`ENOENT` read error instead of masking genuine I/O failures (`EACCES`/`EIO`).

## Verification
- `transports.test.ts` stable across repeated runs (42 passed each).
- No public-API change; no changeset (test-only).

## Context
Extracted from #58 (the engram-server 0.34.0 SDK alignment) so the logger fix can be reviewed/landed independently of the SDK work. Once this merges to `main`, #58 will inherit it on rebase and its duplicate logger commits will drop out.


owenjohnson:
Pushed 619d8ba addressing all 4 findings — these were good catches on my own test.

- **[HIGH] poll budget vs testTimeout** → replaced the hand-rolled `500×10ms` loop (which equalled Vitest's default 5s `testTimeout`) with `vi.waitFor(..., { timeout: 3_000, interval: 10 })` — native retry, proper timeout handling, well clear of the framework timeout.
- **[MED] descriptive failure** → the `vi.waitFor` callback throws `auto-flush incomplete: N/3 lines` (and `log file not created yet`), so a real failure reports *why*, not just `Expected 3, Received 0`.
- **[MED] sync I/O blocking the flush** → switched to `await readFile(...)` (`fs/promises`) so the event loop stays free to run the flush's promise chain between polls.
- **[LOW] JSONL validity** → added a per-line `JSON.parse` check so 3 empty/malformed lines can't pass.

Still does **not** call `flush()` — verifies the auto-flush path fires on its own. Stable across repeated runs; build + lint green.

owenjohnson:
**Scope update:** #58 merged to `main` and carried the initial de-flake (the hand-rolled poll). I've merged `main` into this branch, so #61's diff is now just the **refinement to `vi.waitFor`** — the improvement the review here asked for:

- native `vi.waitFor` (3s budget, clear of the 5s `testTimeout`) instead of the hand-rolled `500×10ms` loop that equalled the framework timeout
- `await readFile` (`fs/promises`) so the event loop stays free for the flush's promise chain
- descriptive in-callback failure (`auto-flush incomplete: N/3 lines`)
- per-line JSONL validity check

Diff vs `main`: `transports.test.ts` only (the `maxBufferSize` test). Stable across repeated runs; build + lint green.

owenjohnson:
Pushed — addressed the 3 LOW nits: `const lines = await vi.waitFor(...)` (callback returns the array, no outer-scope mutation); the per-line check now carries a message (`line is not valid JSON: …`); and a comment clarifies the post-waitFor `toHaveLength(3)` guards against over-write (waitFor only resolves at >= 3). 42 tests green.

owenjohnson:
Applied the MED; rationale on the 3 LOWs.

- **[MED] verify content** → done: each line is parsed AND its `message` asserted against `["Entry 1","Entry 2","Entry 3"]`, so valid-but-wrong JSON fails. (74545c4)

Rationale (not changed):
- **[LOW] `filter(Boolean)` masks blank-line regressions** — this reverses last round's review on this same line, which asked me to *add* `.filter(Boolean)` "for consistency with the rest of the file (line 396)." I'm keeping it consistent with that prior guidance and with the sibling test; the content assertion above (exactly 3 messages, in order) already catches a spurious-blank-line regression (it'd shift/replace an expected message). Flagging the round-over-round reversal rather than oscillating the line.
- **[LOW] connect waitFor `>=3` to `==3`** — the comment directly above the assertion already states "waitFor only resolves once >= 3 are present; this asserts exactly 3 (no over-write)."
- **[LOW] non-ENOENT retried for the full budget** — `vi.waitFor` surfaces the last thrown error on timeout, and the callback already re-throws non-ENOENT with `{ cause }` + path; in a per-test temp dir an `EACCES`/`EIO` is not a realistic path, and the worst case is a ≤5s delay before the (preserved) error surfaces. Not worth a bespoke fail-fast wrapper in a test.

42 tests, lint green. This is a one-test PR; I'd suggest it's complete.

owenjohnson:
Fixed the genuine item; capping the rest.
- **[MED] misleading non-ENOENT comment** → correct: vi.waitFor retries every throw, so non-ENOENT surfaces as the last error when the budget elapses, not instantly. Comment fixed; the unexpected error is wrapped with cause+path.

Rationale (not changed):
- **[LOW] 3000ms vs testTimeout** — already addressed a round ago: the `it(...)` is pinned to `{ timeout: 5_000 }`, so the 3s vi.waitFor no longer relies on the global default. This finding looks stale.
- **[LOW] expect() inside map** — functionally correct (a failed assertion throws out of the map); a parse failure surfaces with the line content. Not worth restructuring.
- **[LOW] only checks `message`, not level/timestamp** — this round asks for *more* structural assertions after last round asked to add content checks; the message+order assertion already catches swap/dup/missing-entry regressions. Diminishing returns.

42 tests, lint green. One-test PR — complete; recommend merge.

owenjohnson:
No code change this round — these re-raise or reverse prior guidance, and there's no new correctness issue. Capping per the convergence note.

- **[MED] non-ENOENT retried for the full budget** → this asks to restore the hand-rolled `if (code !== 'ENOENT') throw` fast-exit loop — i.e. revert the `vi.waitFor` migration an earlier round of this same review required. I'm not oscillating the implementation back and forth. The practical impact is nil: this is a per-test temp dir, EACCES/EIO won't occur, and the worst case is the (cause-preserving) error surfacing after the ≤3s budget rather than instantly. The comment now states this accurately.
- **[LOW] filter(Boolean) masks blank lines** → 3rd time on this exact line, and it reverses the earlier round that asked me to *add* `.filter(Boolean)` "for consistency with line 396." The message+order assertion (`["Entry 1","Entry 2","Entry 3"]`) already catches a stray-blank-line regression (it would shift/replace an expected message).
- **[LOW] expect() inside map / non-object JSON** → functionally correct; a parse failure throws out of the map with the line content.
- **[LOW] close() within the remaining budget** → speculative (confidence: low); `close()` after a resolved `waitFor` is fast, and the `it()` budget is 5s.

This is a one-test PR that's been green for several rounds; the substantive de-flake (the actual goal) is done. Recommend merge.

Issue #62, opened 2026-06-05
Harden: exhaustive runtime response-shape validation across SDK resources


owenjohnson:
## Context

PR #58 (engram-server 0.34.0 alignment) added runtime shape guards to the methods it touched (maintenance `vacuum`/`tombstoneCleanup`/`changelogCompact`, sync `push`/`pushTo`/`pull`/`listConflicts`/peers). Each review round generalized the pattern to one more method.

To **cap** that expansion within #58 (the substantive contract was settled early; remaining items are defensive depth), further shape-hardening is tracked here as a follow-up.

## Proposed follow-up

Apply a consistent runtime response-shape validation pattern across the remaining SDK resources that currently trust the typed envelope without a runtime check — e.g. `getStatus`, `resolveConflict`, `pullFrom`, and the analogous read paths in other resources (gc, entities, terrafirma, export-import, …). Options:

- A small shared `requireData` / `requireShape` helper (extract the one added in `sync.ts`).
- Or a lightweight schema-validation pass at the client boundary.

Goal: a server schema drift surfaces as a structured `EngramError` at the boundary rather than a `TypeError` at the call site, uniformly.

## Out of scope for #58
Doing this exhaustively across ~20 resources would balloon the alignment PR. #58 covers the methods it changed; this issue covers the rest.

🤖 Capped from PR #58 review (rounds 4-6 defensive-depth expansion).

PR #64, opened 2026-06-05
fix(sdk): shape-guard the remaining sync read paths (advances #62)

owenjohnson:
## Summary

Advances #62 (exhaustive runtime response-shape validation). #58 added shape guards to the sync methods it touched (`push`/`pushTo`/`listConflicts`/peers/`pull`); this completes the **sync resource** by guarding the three that were still unguarded:

- `getStatus()` — `requireData` + `schemaVersion` is a string / `changelogSize` is a number.
- `pullFrom()` — `requireData` + `entriesStreamed` / `duration` are numbers.
- `resolveConflict()` — `requireData` + the resolved conflict has a string `id`.

A null or malformed `data` envelope now throws a structured `EngramError` at the boundary instead of a `TypeError` at the call site, reusing the existing `requireData` helper for consistency.

## Scope
This is the **sync-resource increment** of #62. Other resources' read paths can adopt the same pattern incrementally; #62 stays open to track that. Kept deliberately small to stay reviewable.

## Verification
- `npm run build` ✅ · `npm test` ✅ **419 passed** (+3 negative-path tests) · `npm run lint` ✅ · patch changeset added.

### DIFF
--- a/.changeset/sdk-sync-shape-guards.md
+++ b/.changeset/sdk-sync-shape-guards.md
@@ -1,0 +1,9 @@
+---
+"@centient/sdk": patch
+---
+
+Complete runtime response-shape validation across the sync resource:
+`getStatus`, `pullFrom`, and `resolveConflict` now reject a null/malformed
+`data` envelope with a structured `EngramError` (instead of a downstream
+`TypeError`), consistent with the guards already on `push`/`pushTo`/
+`listConflicts`/peers. No change for well-formed responses. (Advances #62.)
--- a/packages/sdk/src/resources/sync.ts
+++ b/packages/sdk/src/resources/sync.ts
@@ -446,7 +446,17 @@ export class SyncResource extends BaseResource {
       "GET",
       "/v1/sync/status"
     );
-    return response.data;
+    const data = requireData(response.data, "GET /v1/sync/status");
+    if (
+      typeof data.schemaVersion !== "string" ||
+      typeof data.changelogSize !== "number"
+    ) {
+      throw new EngramError(
+        "Unexpected GET /v1/sync/status response shape",
+        "INTERNAL_ERROR",
+      );
+    }
+    return data;
   }
 
   /**
@@ -486,7 +496,17 @@ export class SyncResource extends BaseResource {
       "POST",
       path
     );
-    return response.data;
+    const data = requireData(response.data, "POST /v1/sync/pull-from");
+    if (
+      typeof data.entriesStreamed !== "number" ||
+      typeof data.duration !== "number"
+    ) {
+      throw new EngramError(
+        "Unexpected POST /v1/sync/pull-from response shape",
+        "INTERNAL_ERROR",
+      );
+    }
+    return data;
   }
 
   /**
@@ -532,6 +552,13 @@ export class SyncResource extends BaseResource {
       `/v1/sync/conflicts/${encodeURIComponent(id)}/resolve`,
       params
     );
-    return response.data;
+    const data = requireData(response.data, "POST /v1/sync/conflicts/{id}/resolve");
+    if (typeof data.id !== "string") {
+      throw new EngramError(
+        "Unexpected POST /v1/sync/conflicts/{id}/resolve response shape",
+        "INTERNAL_ERROR",
+      );
+    }
+    return data;
   }
 }
--- a/packages/sdk/src/resources/sync.ts
+++ b/packages/sdk/src/resources/sync.ts
@@ -500,13 +500,16 @@ export class SyncResource extends BaseResource {
       path
     );
     const data = requireData(response.data, "POST /v1/sync/pull-from");
+    // Required numeric fields must be present; maxSeq is optional/nullable —
+    // only reject it when present with a non-string value (an absent or null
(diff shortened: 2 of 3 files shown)


owenjohnson:
Round 2 — completed the partial guards (the two HIGHs were valid: a partial guard gives false confidence):
- **[HIGH] getStatus** → validates all 5 `SyncStatus` fields (instanceId/schemaVersion/peersCount/activeLinksCount/changelogSize), not 2.
- **[HIGH/LOW] resolveConflict** → validates the non-nullable `SyncConflict` fields (id, entityType, entityId, fieldName, winner, resolution, createdAt) + presence of localValue/remoteValue/resolvedAt, not just `id`.
- **[MED] pullFrom** → also requires `maxSeq` present (string | null).
- **[MED] messages** → all three now describe the expected shape (matching the `listConflicts` style) for triage.
- **[HIGH test-coverage] getStatus** → added wrong-type and missing-instanceId tests so the `typeof` branches are exercised, not just the null-data path.

`npm run build` ✅ · `npm test` ✅ **421 passed** · `npm run lint` ✅

owenjohnson:
Round 3 — all 6:
- **[HIGH] resolveConflict** → the guard now actually presence-checks `localUpdatedAt`/`remoteUpdatedAt` (the comment claimed it; the code didn't). Genuine fix.
- **[MED] maxSeq** → now `maxSeq === null || typeof maxSeq === 'string'` (was presence-only; `maxSeq: 42` slipped through).
- **[LOW] winner/resolution unions** → validated against their members (`local|remote`, `auto_lww|manual`), not just `typeof string`.
- **[MED] pullFrom coverage** → added null-data and wrong-type(`entriesStreamed`) tests (symmetry with getStatus) + a bad-`maxSeq` test.
- **[LOW] resolveConflict coverage** → added out-of-union winner/resolution tests; the happy fixture now carries `localUpdatedAt`/`remoteUpdatedAt`.

`npm run build` ✅ · `npm test` ✅ **425 passed** · `npm run lint` ✅

owenjohnson:
The two HIGHs were genuine and important — fixed in 164f17d.

- **[HIGH] absent-key rejection (resolveConflict timestamps + pullFrom maxSeq)** → the round-4 guards I added were *too strict*: requiring a nullable field to be `string | null` rejects a response that legitimately **omits** the optional key (absent vs explicit `null`). Loosened both: nullable fields are validated only when present with a non-string value; absent/null pass. Added regression tests asserting an absent `maxSeq` and absent conflict timestamps are accepted. Also dropped the `localValue`/`remoteValue` presence checks (they're `unknown`-typed — nothing useful to assert).

Rationale (not changed):
- **[MED] winner/resolution should be validated against the literal union** → this reverses the round-4 finding (forward-compat) that asked me to use `typeof` instead of the closed set. I'm keeping `typeof` (a future server-side enum value must not throw `INTERNAL_ERROR`). Flagging the round-over-round contradiction.
- The remaining MED/LOW (per-field test symmetry for getStatus/peersCount etc., helper placement) are leaf coverage nits on a guard that's now intentionally minimal.

427 tests, build + lint green. With the over-strictness fixed, I'd consider this complete and ready to merge.

owenjohnson:
The HIGH was genuine and is now fixed at the source (164f17d → this commit).

- **[HIGH] maxSeq could return undefined for a `string | null` type** → instead of more validation, **normalize**: `pullFrom` returns `maxSeq ?? null` and `resolveConflict` returns the three nullable timestamps `?? null`. The return now always matches the declared types, and absent keys are accepted. Tests assert the normalized `null`. This deliberately ends the maxSeq/timestamp oscillation across the last three rounds (reject-absent → "too strict" → "undefined leaks") with one stable rule.
- **[MED] resolveConflict null-data test** → added.

Capping the rest (rationale):
- **[MED type-safety] winner/resolution accept any string vs the literal union** — this is the *direct reverse* of the round-4 finding that told me to drop the closed-set check for forward-compat. I'm keeping `typeof` so a future server enum value doesn't throw; widening the interface type to `string` would lose useful caller-side typing. Flagging the round-over-round contradiction rather than oscillating again.
- **[LOW] error message / [LOW] helper placement / [LOW] per-field test symmetry** — leaf nitpicks on a guard that's now intentionally minimal + normalizing; no correctness impact.

428 tests, build + lint green. This PR (advances #62) is complete and correct — recommend merge.

owenjohnson:
Capping — no genuine correctness bug; this PR (advances #62) is complete.
- **[MED] winner/resolution typeof vs literal union** → re-raise of a point this review *reversed itself on*: an earlier round flagged the closed-set check as forward-incompatible and had me switch to `typeof`. Keeping `typeof` so a future server-side enum value doesn't throw `INTERNAL_ERROR`. The forward-compat trade-off is documented inline.
- **[MED/LOW] test isolation (id-only fixture, duration wrong-type, peersCount: null)** → test-symmetry refinements; each guard is a simple independent `||` branch and the happy + representative negative paths are covered. No correctness impact.
- **[LOW] NaN/Infinity pass typeof number** → JSON can't encode `NaN`/`Infinity` (they serialize to `null`, which the guard already rejects), so this path isn't reachable from a real response.

428 tests, build + lint green. Recommend merge; broader shape-validation rollout is tracked in #62.

Issue #11, opened 2026-04-04
Require positive human auth for vault unlock


owenjohnson:
## Context

From the threat model (Adversary Class 4): an autonomous agent with no human present could access vault secrets because human-presence detection relies on environment variable checks, not actual biometric confirmation.

## Current state

`@centient/secrets` unlocks the vault via macOS Keychain (Touch ID / password). However, the check for whether a human is actually present is based on environment variables (e.g., session indicators). An agent running in a long-lived session could pass these checks without a human at the keyboard.

## Desired behavior

- Each sensitive vault operation (get, set, delete) should require a fresh biometric/keychain unlock token
- Token should have a short TTL (e.g., 5 minutes) tied to an actual Keychain access event
- Agents running autonomously (no human) should fail the check gracefully with a clear error
- Configurable policy: strict (always require), session (require once per session), permissive (current behavior)

## References

- `support/standards/threat-model.md` — Adversary Class 4
- `packages/secrets/` — current vault implementation
- Related: centient-sdk#8 (1Password backend), centient-sdk#9 (audit logging)

owenjohnson:
## Design direction — positive human auth, scoped to the entry point

Reframing the issue after working through the maintainer's architecture: the goal isn't "biometric everywhere." It's **positive human auth appropriate to each entry point**, plus least-privilege key distribution. A password is fine for some entry points and useless for others — what matters is *who/what can satisfy the unlock at that point* and *what holds the key afterward*.

### Two entry points, two providers

| Entry point | Who unlocks | Right factor |
|---|---|---|
| Interactive dev on macOS running `centient secrets get` directly | human at the keyboard | **keychain / Touch ID** (presence — software can't forge it) |
| **Maintainer multi-container harness** | human runs `maintainer start` once | **passphrase** entered once, KDF-derived key held in the deterministic broker |

The maintainer case is the one the current providers (`keychain`, `1password`) don't serve well: it's a headless/Linux host with no Touch ID and possibly no desktop keychain. The architecture that makes a passphrase sound there:

- The **maintainer CLI** (human-invoked) spawns a **deterministic, non-agent harness** that holds the vault key in memory.
- That broker passes **scoped, least-privilege keys** to **container harnesses**, which run agents **inside containers**.
- Agents never hold the vault and never see the unlock secret. **The container is the trust boundary** (already the threat model's listed mitigation).

[...]

Issue #37, opened 2026-04-19
centient secrets set truncates multi-line values at first newline


owenjohnson:
## Observed

Running the following in sequence produced a truncated stored value:

\`\`\`
centient secrets set github-app-private-key
# interactive paste of a GitHub App PEM (multi-line)

centient secrets get github-app-private-key | wc -l
# 1
# The single line is just "-----BEGIN RSA PRIVATE KEY-----" — the RSA header only.
\`\`\`

The PEM in \`~/.config/maintainer/config.json\` is the full multi-line key. The value stored in the centient vault is truncated to the first line of the paste.

## Root cause

\`packages/secrets/src/cli/secrets-cli.ts\` → \`prompt(message, hidden=true)\` (compiled form visible at \`dist/cli/secrets-cli.js:103-147\`).

In hidden mode it enables raw TTY mode and listens for \`data\` events, resolving on the first \`\\n\` or \`\\r\`:

\`\`\`js
const onData = (char) => {
    if (char === \"\\n\" || char === \"\\r\") {
        stdin.setRawMode?.(wasRaw || false);
        stdin.pause();
        stdin.removeListener(\"data\", onData);
        process.stdout.write(\"\\n\");
        rl.close();
        resolve(input);   // ← input is whatever came before the first newline
    }
    ...
};
\`\`\`

For an interactive paste, the terminal delivers the pasted content character-by-character (or in chunks) and the **first** \`\\n\` between the header line and the key body terminates input. Only \`-----BEGIN RSA PRIVATE KEY-----\` makes it into \`input\` before \`resolve\` fires.

Same handler likely misbehaves under a non-TTY stdin pipe too (chunk boundaries make the \`char === \"\\n\"\` branch firing behavior non-deterministic), though that's a separate concern.

## Why it matters

Multi-line secrets are a real use case — RSA/EC private keys are the obvious one, and that's exactly what bit me. The symptom is silent: set succeeds, get returns something plausible-looking, and the truncated value only breaks later when something downstream tries to use it as a valid PEM.

Specifically, this blocked the maintainer daemon's credential migration workflow: any setup flow that says \"paste your PEM into \`centient secrets set\`\" has been silently storing a 30-character header string instead of the 1700-character key.

## Reproduction

\`\`\`
# Create a multi-line value
cat <<EOF' > /tmp/multi.txt
line one
line two
line three
EOF'

# Paste the contents into the prompt
centient secrets set test-multiline
# (paste all three lines)

[...]

PR #67, opened 2026-06-07
feat(secrets): passphrase key-provider for headless vault unlock

owenjohnson:
## Summary

- Adds a `passphrase` KeyProvider for `@centient/secrets` as Step 1 of #11.
- Derives the vault key from a typed passphrase plus per-vault salt instead of storing the vault key.
- Wires provider resolution, package exports, `openVault` diagnostics, and CLI init setup for derivation-based providers.

## Resolved decisions

- KDF: Node `crypto.scryptSync`, zero new dependency. Production params: `N=131072`, `r=8`, `p=1`, `keyLength=32`, `maxmem=192 MiB`.
- Metadata location: `vault.passphrase.json` beside the vault file. The file stores only version, KDF name, KDF params, salt, and an HMAC-SHA256 verifier.
- Non-TTY behavior: hard fail closed with a clear error. No env var, file, or stdin override is added in this step.
- Policy modes: provider only in this PR. strict/session/permissive policy wiring remains a follow-up.

## Security notes

- The passphrase and derived key are never persisted to env, file, keychain, or 1Password.
- `storeKey()` refuses caller-supplied random keys for the passphrase provider; `setupKey()` derives the bootstrap key and persists only metadata.
- Wrong passphrases fail against the verifier before returning a garbage key to vault decrypt.
- Metadata is written with mode `0600` and `deleteKey()` removes the salt/verifier file.

## Follow-ups

- Policy modes and TTL hooks through `vault/policy.ts`.
- Audit unlock and denied-unlock wiring from #9.
- Native Touch ID provider for macOS interactive presence.

## Verification

From `packages/secrets`:

- `npm run build`
- `npm test`
- `npm run lint`

Refs #11.


owenjohnson:
## Code review — 10-dimension parallel review (substituting for the failed automated review, see #68)

**Verdict: remediate-then-merge.** No critical findings. Crypto design is sound; secrets package tests pass 202/202. Findings below; the 4 HIGH items and 2 security MEDIUMs are being addressed on this branch.

### Verified solid
- **KDF design**: scrypt N=2^17/r=8/p=1 (~128MB, OWASP-aligned), 32-byte CSPRNG salt, HMAC-SHA256 verifier with domain separation compared via `timingSafeEqual`; `validateParams` blocks weak-KDF downgrade via tampered metadata.
- **Zero new dependencies** — pure `node:crypto`, fully available on Node >= 20.
- **Backward compatible**: KeyProvider changes are additive-optional; passphrase appended as the *last* auto-detect fallback (existing keychain/1Password users resolve identically); `minor` changeset correct.
- **No silent degradation**: `getLastError()` plumbed through `openVault` and CLI; non-TTY fails closed; keys zeroed on failure paths; metadata 0600.

### High priority
1. **Test coverage** — `secrets-cli.ts:283-345`: the reworked `initVault()` setupKey/storeKey branch (the PR's core CLI integration) has zero test coverage; same for the `migrateProvider("passphrase")` refusal guard (`:780-787`).
2. **Docs** — `packages/secrets/README.md`: Key Providers table and intro omit `PassphraseProvider`; no security-tradeoff documentation (passphrase strength is the security ceiling; the verifier sidecar enables offline brute-force if exfiltrated; TTY-required/fail-closed; sidecar location/permissions).

### Medium — security
- `hidden-prompt.ts:52-89` — no SIGINT/SIGTERM handler: an external `kill` during the blocking `readSync` skips `finally`, leaving the terminal in raw mode with echo disabled.
- `hidden-prompt.ts:65` — the `Buffer.allocUnsafe(256)` read buffer holding passphrase bytes is never zeroed; the passphrase also travels as an immutable JS string (unwipeable), undercutting the `key.fill(0)` hygiene elsewhere. Minimum: zero the buffer in `finally`, document the string-residue limitation.

### Medium — other (follow-up material)
- `scryptSync` blocks the event loop during unlock; the sync `getKey()` contract leaves no async escape — document for daemon callers of `openVault()`.
- Layering: a library provider (`key-providers/`) imports from `src/cli/` — move the prompt seam to a neutral module.
- Migrate asymmetry: migrate-TO-passphrase is advertised in usage/validTypes then hard-rejected with no issue reference (migrate-FROM works); reference #11 or gate both directions.
- Untested unlock-time branches: prompt interruption, KDF failure at unlock, EOF/Ctrl-C in hidden-prompt, unicode/empty passphrase round-trips.
- KDF default params lack a rationale comment; CLI header still says "prompts biometric".

### Low (selected)
- Non-atomic metadata write (`writeFileSync`, no temp+rename) — crash mid-write = vault lockout; `session-vault.ts` already has the atomic pattern.
- `secrets status` calls `getKey()` to test availability — full interactive prompt + 128MB scrypt to print a status line; use a metadata existence check.
- Magic literals 32/16 in validators; bare-string error codes (suggest a `PassphraseErrorCode` union).
- Passphrase provider offers no human-presence guarantee (vs Touch ID) — relevant to the #11 policy work; document.

### Pre-existing (not this PR)
- `dbus-next` transitive tree: 12 audit advisories (1 critical, 6 high) via `usocket > node-gyp > request` — deserves a separate `type/chore` P2 issue.
- `packages/secrets` declares `engines: >=18` vs repo standard `>=20`.

---
*Review: 10 parallel dimension agents (security, performance, quality, tests, architecture, docs, standards, logging, dependencies, completeness) over `origin/main...HEAD`, 11 files, +1198/-51.*

owenjohnson:
Remediation pushed in 5cbb4d7 addressing the review's 4 HIGH findings + 2 security MEDIUMs:

- **hidden-prompt signal safety**: one-shot SIGINT/SIGTERM/SIGHUP handlers restore raw mode/bracketed paste and re-raise, so an external `kill` mid-prompt no longer corrupts the terminal.
- **Passphrase buffer hygiene**: the transient read buffer is zeroed in `finally`; the JS-string residue limitation is documented on `promptHiddenSync`.
- **README**: `PassphraseProvider` added to the Key Providers table + a security-tradeoffs section (offline brute-force via the sidecar, no human-presence guarantee, TTY-required/fail-closed, event-loop-blocking unlock guidance for daemons).
- **CLI test coverage**: new `secrets-cli.init-migrate.test.ts` covers the `initVault` setupKey/storeKey branches (success, failure with `getLastError` surfacing, key zeroing, no vault write on abort) and the migrate-to-passphrase refusal guard; two new hidden-prompt tests pin signal-handler install/removal.

Tests: 210/210 pass, `tsc` clean. Remaining medium/low findings from the review are follow-up material (layering of the prompt seam, migrate asymmetry issue ref, unlock-time error-branch tests, KDF rationale comment, stale 'biometric' CLI wording).

owenjohnson (commented):
Verified: Read PR body and full diff; checked passphrase provider, CLI integration, hidden prompt signal handling, and tests.
Findings: 1. packages/secrets/src/cli/hidden-prompt.ts:96 — HIGH: external SIGINT/SIGTERM/SIGHUP handling can recurse instead of terminating cleanly. The process.once handler is still installed when it calls process.kill(process.pid, signal), so the re-raised signal invokes the same handler again before the default disposition is restored. Remove the handler before re-sending the signal, or set a guard and re-raise after cleanup, with a focused test.

owenjohnson:
## Response to codex CHANGES_REQUESTED + maintainer R19 — addressed in 895b88d

Both findings were probed empirically (child process + external `kill`) before changing anything.

### codex HIGH — "signal handler can recurse via process.kill re-raise" (hidden-prompt.ts:96)

**Rebutted as described, but the probe found a worse, real bug on the same path — fixed.**

The recursion as described cannot occur: `process.once` removes the wrapper *before* invoking the handler, so by the time `process.kill(process.pid, signal)` runs there is no listener for that signal and the re-raise hits the default disposition. Deeper: the prompt is fully synchronous, so the JS handler can never run *during* the prompt at all — libuv queues the signal and can only dispatch it after the prompt unwinds.

That timing is exactly where the real bug was: the `finally` block removed the listeners synchronously, before the event loop could dispatch the queued signal — so the queued dispatch was dropped and **the process silently ignored an external SIGTERM received during the prompt**. Probe: child blocked in the prompt survived `kill -TERM`, completed the prompt, and kept running. The comment block's claimed "restore and re-raise" was dead code.

Fix in 895b88d:
- listener removal is deferred one tick (`setImmediate`), so a signal queued during the prompt dispatches and re-raises with default disposition;
- the handler removes **all** prompt listeners before re-raising (codex's recommendation — now load-bearing, and it also prevents stale sibling-signal handlers);
- `restoreTerminal` is idempotent (guards the double-restore on the deferred path);
- focused child-process test asserts defer-then-re-raise end-to-end: child survives SIGTERM while prompting (terminal protected), completes the read, then terminates with `signal=SIGTERM`; a `SIGNAL_SWALLOWED` sentinel guards against regression.

### maintainer R19 HIGH — "readSync in while(true) without bounds"

**Half rebutted, half fixed.** Blocking indefinitely waiting for keystrokes is the contract of an interactive prompt — a timeout would be a UX-breaking change with no security gain, and the no-bytes case (`bytesRead === 0`) already returns null. Rebutted that part. The valid kernel is unbounded *accumulation*: a hostile or broken input source feeding endless bytes with no submit grew `state.input` without bound. Fixed: accumulation is capped at `MAX_HIDDEN_INPUT_LENGTH` (64 KiB — multi-line PEM/config pastes per #37 fit comfortably); exceeding it drops the buffered reference and aborts with `HiddenPromptError`. Boundary tests added (exactly-64KiB accepted, beyond rejected with terminal restored).

`make check` green: tsc 0 errors, 1242 passed / 14 skipped.

owenjohnson:
## Response to maintainer round (request fab46839) — both new findings rebutted, no code change

**[HIGH] "process.kill in the handler re-triggers the same signal before the handler completes — defer with setImmediate": rebutted, twice over.**

1. The first statement in `onSignal` is `removeSignalHandlers()`, which removes **all** prompt listeners before the re-raise. By the time `process.kill(process.pid, signal)` executes, no JS listener exists for that signal, so the kernel-default disposition applies and the process terminates. There is no handler left to recurse into.
2. Even with a listener installed, Node signal dispatch cannot synchronously re-enter a JS handler: the libuv C handler writes to a self-pipe and the JS callback is dispatched on a later event-loop iteration. Synchronous recursion is structurally impossible in Node's signal model — "undefined behavior" does not apply.

This is verified empirically in the branch: the child-process test delivers a real external SIGTERM mid-prompt and asserts the process terminates cleanly with `signal=SIGTERM` after the prompt completes — no recursion, no hang. Deferring the re-raise with `setImmediate` would be strictly worse: it reopens the window the current code just closed, letting arbitrary code run (or the process exit some other way) between handler completion and the deferred kill, i.e. exactly the swallowed-signal bug 895b88d fixed.

**[MEDIUM] "readSync can return -1 on EAGAIN/EINTR, so subarray(0, -1) misbehaves": rebutted — that return value cannot occur.**

`fs.readSync` follows the sync fs API contract: it returns a non-negative byte count and **throws** on error; it never returns -1. Verified directly: `readSync(badFd, ...)` throws `EBADF` (and the suite's existing "restores raw mode when fd reads fail" test exercises precisely this thrown path — terminal restored, listeners disarmed). EINTR is retried inside libuv's fs layer; a hypothetical EAGAIN (non-blocking fd) surfaces as a thrown exception, which propagates after the `finally` cleanup — the prompt fails closed with the terminal restored, per the documented contract. The `bytesRead === 0` check is EOF handling, not error handling, and `buffer.subarray(0, -1)` is unreachable.

owenjohnson:
Coordinator note (2026-06-11 ~16:35Z): operator has discontinued codex reviews — mbot (centient-maintainer[bot]) is the sole review gate again. The 2026-06-10 codex CHANGES_REQUESTED no longer blocks; work to mbot's review rounds toward an APPROVED on current head.

owenjohnson:
Branch refreshed onto current main (merge, no force-push) — the full hardening train (#71–#77, #79) is now included; no file overlap with this PR's secrets-package changes. `make check` green on the merged branch: tsc 0 errors, full workspace suite passing, claudemd-check OK.

Standing state of the two findings from the last round, both rebutted upthread with empirical evidence and unchanged by this refresh:
1. **Signal-handler re-raise (HIGH)** — `onSignal` removes all prompt listeners before `process.kill`, so the re-raise hits the kernel-default disposition; Node dispatches signals via libuv's self-pipe, making synchronous handler re-entry structurally impossible. The branch's child-process test demonstrates clean `signal=SIGTERM` termination. Deferring the re-raise via `setImmediate` would reintroduce the swallowed-signal window 895b88d closed.
2. **`readSync` returning −1 (MEDIUM)** — sync fs APIs throw on error and never return negative; the thrown path is exercised by the existing failing-fd test (terminal restored, listeners disarmed). `subarray(0, -1)` is unreachable.

owenjohnson (commented):
🤖 codex-review — VERDICT: LGTM

Verified:
- Read PR body, current head 83018b77f1f2e833f2541da600ac8c7d5d6b6885, review history, full diff, and current-head contents for passphrase-provider.ts, hidden-prompt.ts, resolve.ts, types.ts, session-vault.ts, secrets-cli.ts, and hidden-prompt tests.
- Checked passphrase metadata validation, scrypt parameter validation, verifier handling, non-TTY provider detection, CLI init key zeroing, migration guard, signal-handler cleanup behavior, and focused tests.
Findings: none.

PR #72, opened 2026-06-11
docs(monorepo): fix CLAUDE.md/CHANGELOG drift and harden claudemd-check [gap-none-docs-drift]

owenjohnson:
Hardening backlog ticket gap-none-docs-drift (docs/hardening/BACKLOG.md) — phase 5 of the hardening pipeline.

## Gap

Three doc/contract drifts: CLAUDE.md version table stale (events 0.2.3, logger 0.17.1, wal 0.3.3); sdk-python CHANGELOG.md:29-30 claims entities/extraction "not yet implemented, planned v1.1.0" though they shipped in 1.0.0; CLAUDE.md says "20 resources", actual exports 24+.

## What changed

[...]

PR #74, opened 2026-06-11
chore(monorepo): record npm provenance posture and harden make publish gate [gap-none-publish-provenance]

owenjohnson:
Hardening backlog ticket gap-none-publish-provenance (docs/hardening/BACKLOG.md) — phase 5 of the hardening pipeline.

## Gap

npm provenance attestation was lost when GH Actions was archived (cef5ad7): the old workflow set NPM_CONFIG_PROVENANCE=true; manual `make publish` does not, so public @centient/* packages ship unattested.

## What changed

Spec branch taken: outcome 1 (local provenance NOT supported). npm provenance generation requires a supported CI OIDC provider (GitHub Actions / GitLab CI); a local `npm publish` (which `pnpm changeset publish` shells out to) has no OIDC identity to attest, regardless of org npm settings. Spec branch 2 (minimal workflow_dispatch publish job) would require org-side provisioning (repo NPM_TOKEN secret or npm trusted-publisher config) that cannot be created or verified from this repo and would partially reverse the deliberate cef5ad7 archival — conservative reading taken and recorded; full CI was NOT reintroduced (guard honored), no tokens added to the repo (guard honored), changesets flow untouched (guard honored).

Changes (commit cf958a6):
1. /Makefile — (a) `check` now records a tree fingerprint ({ git rev-parse HEAD; git status --porcelain; git diff HEAD; } | git hash-object --stdin) in .logs/.check-stamp; (b) the `publish` recipe's first action recomputes the fingerprint and FAILS (exit 1, not warn) on missing/mismatched stamp, so publish cannot reach npm without build+check having passed against the exact tree being published — non-bypassable beyond the plain prerequisite (covers `make -o check publish`, `make --touch`); `publish: build check` prerequisite wiring retained; (c) `changeset publish` step now runs with NPM_CONFIG_PROVENANCE=false explicitly, with an in-recipe comment citing cef5ad7, the archived workflow, and a DEFERRED note to restore attestation when a CI publisher returns; (d) Makefile changelog comment appended per file convention.
2. /RELEASING.md — new "What gates a publish (post-CI-archival)" section enumerating the exact gates (build, check, tree-fingerprint stamp, npm auth preflight) and forbidding direct `pnpm changeset publish` outside recovery; new "Provenance" section documenting the limitation, the explicit NPM_CONFIG_PROVENANCE=false choice, and the deferred restoration path (workflow_dispatch + id-token:write + tag-push); recovery and OTP-fallback commands updated to carry NPM_CONFIG_PROVENANCE=false and scoped to resuming a run in which build+check already passed.

Verification: `make --dry-run publish` confirmed recipe expansion/ordering (build → lint/test → stamp → assertion → whoami → changeset version/publish); the stamp assertion was replicated verbatim in shell and tested: no-stamp → REFUSED, stamp+same-tree → PASSED, tree-modified-after-check → REFUSED, reverted-to-checked-tree → PASSED. A live `make publish` bypass test was correctly denied by the sandbox (real publish/push to main) and was not attempted further. `make check` passed from the worktree root (output above, verbatim tail).

Changeset added: no — only Makefile and RELEASING.md changed; no published package code under packages/* was touched, so per repo convention no changeset is required.

Note: this worktree required `git submodule update --init` for scripts/release-toolkit (pinned sha b55a0f9, unchanged) before make targets could run.

## Verification

- make check green: true
- adversarial refutation verdict: survived

Refuter summary: "Not refuted. The branch (cf958a6, Makefile + RELEASING.md only) implements outcome 1 of the spec: provenance is explicitly declined (NPM_CONFIG_PROVENANCE=false on the changeset publish line — verified the env var maps to npm's provenance config), the local-publish limitation and a deferred restore-via-CI note are documented in RELEASING.md, and `make publish` is hard-gated on `check` via a tree-fingerprint stamp that I functionally exercised: valid stamp passes; missing stamp, tracked-file tamper, and new untracked files all abort with exit 1 before any npm command, while plain `make publish` always re-runs build+check as prerequisites. All four guards hold (no CI reintroduced, RELEASING.md accurate, tracked .npmrc empty/no tokens, changesets flow intact), the full suite passes (1219 tests, tsc clean), and origin/main demonstrably lacks the gate. Two non-refuting weaknesses: the fingerprint does not cover the content of already-untracked files (demonstrated; only reachable via deliberate `make -o check`), and the documented OTP-recovery flow bypasses the Makefile gates with only prose constraining it."

## Guards

- [x] Do not reintroduce full GitHub CI (the archival was a deliberate org decision — cef5ad7); only the publish/attestation step is in scope
- [x] RELEASING.md stays accurate for both outcomes
- [x] No npm tokens land in the repo
- [x] compatibility with changesets flow preserved


owenjohnson (commented):

Verified: current head cf958a6. Reviewed PR body, full diff, Makefile publish/check gate, and RELEASING.md provenance/publish-gate documentation.
Findings:
- MEDIUM Makefile:37 — TREE_FINGERPRINT does not actually fingerprint the exact tree state it claims. `git status --porcelain` records untracked file names but not their contents, and `git diff HEAD` ignores untracked content. Since publish later runs `git add -A`, an already-untracked changeset/source file can be modified after `make check` without changing the stamp, then still be included in the version/publish commit. Either include untracked file contents in the fingerprint or reject untracked files during the stamped publish path; otherwise the “check passed against this exact tree” gate and RELEASING.md guarantee are false for a valid dirty worktree.

owenjohnson:
## Note on the remaining LOW finding — rebutted, no change needed

The finding assumes a silent stamp-write failure could **bypass** the publish gate. The gate fails closed in every variant of that scenario:

- The publish gate refuses unless the stamp's content **equals** the current tree fingerprint. A missing, empty, truncated, or corrupted stamp produces a mismatch → `publish` refuses (`[ -z "$stamp" ]` or inequality). A failed write can only cause a spurious refusal, never a bypass — bypass would require the write to *succeed* with the correct fingerprint, which is the intended path.
- The write cannot fail silently anyway: the recipe is `mkdir -p .logs && fingerprint > stamp` with no `-` prefix and no `|| true`, so a failed `mkdir` or failed redirection exits non-zero and make fails the `check` target loudly.

Verifying the write after the fact would re-check a condition whose failure already fails the target, to guard a gate that already fails closed. Leaving as-is.

owenjohnson:
Coordinator note (2026-06-11 ~16:35Z): operator has discontinued codex reviews — mbot (centient-maintainer[bot]) is the sole review gate again. Outstanding 🤖 codex-review verdicts on this PR no longer block; work to mbot's review rounds. Merge rule: mbot APPROVED on current head ⇒ coordinator admin-merge. This PR is currently conflict-DIRTY after sibling merges — rebase when ready (rebase voids prior approvals; mbot will re-review).

owenjohnson:
Rebased-by-merge onto current main (no force-push per repo policy): resolved the Makefile conflict by combining #72's `claudemd-check` gate dependency with this PR's fingerprint stamp — `check` now runs lint + test + claudemd-check AND records the stamp. Also carries the same jitter-test `text()` mock fix as #78 (main is currently red without it; identical content merges cleanly). `make check` green on the branch: 470/470 sdk tests, stamp written, claudemd-check OK.

PR #71, opened 2026-06-11
fix(sdk): add jitter to retry backoff [gap-none-retry-jitter]

owenjohnson:
Hardening backlog ticket gap-none-retry-jitter (docs/hardening/BACKLOG.md) — phase 5 of the hardening pipeline.

## Gap

All 6+ retry sites in packages/sdk/src/client.ts (470, 489, 558, 595, 665, 748) sleep exactly retryDelay*attempt with zero jitter — synchronized consumers retry in lockstep against a struggling engram-server.

## What changed

Implemented jittered retry backoff per spec. Commit a1b07c1 on remediate/gap-none-retry-jitter (pushed).

Changes:
- /packages/sdk/src/client.ts: added private backoffDelay(attempt) returning `this.retryDelay * attempt + Math.random() * this.retryDelay * 0.5` (single line, full-jitter-lite, linear base preserved); replaced ALL inline `this.sleep(this.retryDelay * attempt)` calls with `this.sleep(this.backoffDelay(attempt))` — there were 8 retry sites, not 6 (ticket said "6+"): _requestRaw (5xx + network), _requestRawBody (5xx + network), _requestFormData (5xx + network), request (5xx-EngramError + network). Added JSDoc on the private retryDelay field and on backoffDelay documenting the jitter and worst-case bound.
- /packages/sdk/src/types.ts: EngramClientConfig.retryDelay JSDoc updated to document the jitter and that worst-case total retry time is the linear budget + 0.5*retryDelay per attempt.
- /packages/sdk/tests/client.test.ts: new "Retry Backoff Jitter" describe — (1) stubbed Math.random pins exact lower bound (random=0 → exactly base, budget unchanged), midpoint, and near-upper bound strictly < base + 0.5*retryDelay for attempts 1,2,3,5; (2) real-Math.random property test asserts [base, base + 0.5*retryDelay) over 50 samples per attempt; (3) end-to-end retry path test spies the private sleep and asserts the jittered argument (12.5 for retryDelay=10, attempt=1, random=0.5); (4) source-grep enforcement test reads src/client.ts and asserts every this.sleep(...) call is exactly this.sleep(this.backoffDelay(attempt)) (>=8 sites) and `retryDelay *` appears on exactly one line (backoffDelay's return). Existing retry-count tests unchanged; TimeoutError/EngramError/NetworkError non-retry semantics untouched (only sleep arguments changed).
- Changeset added: yes — /.changeset/remediate-gap-none-retry-jitter.md, "@centient/sdk": patch.

[...]

PR #75, opened 2026-06-11
feat(sdk): optional client logger for retries, timeouts and exhaustion [gap-none-sdk-client-logging]

owenjohnson:
Hardening backlog ticket gap-none-sdk-client-logging (docs/hardening/BACKLOG.md) — phase 5 of the hardening pipeline.

## Gap

packages/sdk/src/client.ts performs zero logging — retries, timeouts, and network errors are invisible to consumers diagnosing hangs vs retry storms (wal/events/logger packages all instrument; the client is the blind spot).

## What changed

[...]

PR #78, opened 2026-06-11
test(sdk): fix red main — jitter test mocks need text() after #76

owenjohnson:
**Main is currently red** — `make check` fails on `tests/client.test.ts > Retry Backoff Jitter > retry path sleeps the jittered duration from backoffDelay` with `NetworkError: Failed to GET /v1/health: response.text is not a function`. Reproduced on a clean `origin/main` checkout (f93ecdb).

Cause: a semantic conflict in the merge train. #71 (merged first) added a jitter test with inline fetch mocks exposing only `json()`; #76 (forked before #71 landed, merged after) rewrote `request()` to read response bodies via `response.text()`. Each PR was green on its own branch; their union fails.

Fix: the inline mocks now expose `text()` alongside `json()`, matching the shared `mockFetchResponse` helper that #76 already updated. Test-only change.

Verified on this branch: `make check` green — tsc 0 errors, 453/453 sdk tests (1236 total workspace), claudemd-check OK.

**Recommend merging this before anything else** — the maintainer bot's in-container `make check` will fail every review round on every PR until main is green (this also blocks re-reviews of #67/#71 responses and the refreshed #74/#75).

### DIFF
--- a/packages/sdk/tests/client.test.ts
+++ b/packages/sdk/tests/client.test.ts
@@ -1163,17 +1163,20 @@ describe("EngramClient", () => {
       mockFetch = vi.fn().mockImplementation(() => {
         callCount++;
         if (callCount < 2) {
+          const body = { code: "INTERNAL_ERROR", message: "boom" };
           return Promise.resolve({
             ok: false,
             status: 500,
-            json: () =>
-              Promise.resolve({ code: "INTERNAL_ERROR", message: "boom" }),
+            json: () => Promise.resolve(body),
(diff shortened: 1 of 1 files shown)


owenjohnson:
Redundant now: the identical mock fix landed on main inside #75's merge commit (4b55372 → dadfafe), and #74 carried the same content. Verified on current origin/main: the previously-failing jitter test and the full client + logging suites pass (73/73 at the collision point) — main is green again. Closing without merge; branch deleted.

PR #90, opened 2026-06-13
feat(config-loader): add @centient/config-loader package

owenjohnson:
## Summary

New package **`@centient/config-loader`** (`packages/config-loader`, starts at `0.0.0`): layered configuration resolution with caching, write-back, warnings, tilde-app path helpers, and project-root discovery. Zero runtime dependencies; injectable `fs`/`env`/`homeDir`/`cwd`/`logger` seams.

Precedence, highest to lowest: **env > project file > user file > defaults**. Each resolved value carries its source (`getResolved`) for diagnostics and write-back decisions.

This package **resolves and layers only** — it does **not** validate domain shapes. Validation schemas stay in consumers; per-key `EnvCoercer`s are the single opt-in seam where an env string is shaped before it participates in precedence.

## Seeds surveyed

- `platform/centient/packages/centient/src/utils/config.ts` (~1,485 lines) — the main seed. Extracted the generic core (layering, `${VAR}`/`${VAR:-default}` expansion, walk-up, write-back, perms); dropped all centient-specific keys (`engram*`, `metaKnowledge`, memory-extraction config, etc.). The stated `platform/centient/utils/config.ts` path did not exist; located the real file under `packages/centient/src/`.
- `private-packages/touchstone/src/store/discovery.ts` — the walk-up-to-marker root-resolution algorithm, generalised to an injected `FileSystem` and to report the located config file directly.
- `platform/soma/packages/soma/src/runtime/settings.ts` — validate-at-edge env readers; informed the throwing-`EnvCoercer` contract (malformed env override is a hard error, not a silent fallback).

## Design notes

- **No silent fallthrough** (no-silent-degradation): a config file that exists but cannot be parsed (or is a non-object) is a `ConfigError`, never skipped to the next layer. A *missing* file is a legitimate empty layer. A throwing coercer is a hard `ConfigError(INVALID_ENV)`.
- **Observable architecture**: no `Date.now()`/`Math.random()` in core; `fs`, `env`, `homeDir`, `cwd`, and the warning sink are all injectable.
- **Dotted keyspace**: defaults use flat dotted keys; files use nested JSON, flattened before layering so a default and a nested file value for the same logical key always collide correctly. Write-back unflattens back to human-friendly nested JSON.
- **Permissions enforced**: app home created/tightened to `0o700` (`ensureAppHome`), user config written `0o600`.

## Test evidence

`make check` green: lint (tsc 0 errors) + full suite (**1329 passed, 14 skipped**) + claudemd-check (`OK @centient/config-loader (0.0.0)`, all rows match, resource count 31).

The package ships **33 tests** across two files:
- **Precedence matrix** — default < user < project < env, with provenance assertions; bound-keys-only env reads; empty-env-var falls through.
- **Env coercion** — typed coercion + `ConfigError(INVALID_ENV)` on a throwing coercer.
- **Env-ref expansion** — `${VAR}` / `${VAR:-default}` in file string leaves.
- **Malformed-file surfacing** — invalid JSON and non-object files throw `ConfigError(MALFORMED_FILE)`; missing file does not throw.
- **Write-back round-trip** — persists + re-reads through a fresh loader, preserves unrelated keys, refuses to overwrite a malformed existing file.
- **0o600 file / 0o700 home** enforcement (create + tighten-existing).
- **Caching/reload**, **home-env override** with tilde expansion.
- **Walk-up discovery** — config-in-parent, start-dir wins, marker fallback, workspaces `package.json`, null when nothing matches, non-workspaces `package.json` ignored.

## Versioning

`.changeset/config-loader-initial.md` declares a **minor** bump for the new package (`0.0.0`); CLAUDE.md package table row added (version column `0.0.0`, claudemd-check enforced).

## Followups (consumer migrations to queue)

- **centient** — migrate `packages/centient/src/utils/config.ts` to consume `@centient/config-loader` for the generic layering/expansion/discovery/write-back core, keeping its domain keys + `validateConfigSchema` in-repo.
- **engram-server** — replace its ~1,192-line `utils/config.ts` with the shared loader.
- **soma** — adopt the loader for `runtime/settings.ts` env resolution (env bindings + coercers), keeping `OrchestrationSettings` shape validation at the edge.


owenjohnson:
Addressed all 5 findings from the latest maintainer review (commit 4bdee9c). `make check` green (lint + 42 config-loader tests, was 33 + 9 new + claudemd-check). No new runtime deps.

- **[MEDIUM] test-coverage — ConfigLogger warning forwarding** (`tests/loader.test.ts`): added a test that injects a mock `ConfigLogger`, triggers a real warning (malformed `package.json` on the walk-up path), and asserts every collected warning reaches `logger.warn` with `source` context — plus a negative test that the logger is not called when resolution is clean. This also turns the previously-dead logger-forwarding loop into a covered path.

- **[LOW] error-handling — silent catch in `discovery.ts`**: `hasWorkspacesPackageJson` no longer swallows every error in one bare `catch`. It now differentiates a read failure (e.g. permissions) from a JSON parse failure and reports each via a new optional `onWarn` / `DiscoveryWarning` channel, which the loader threads into its non-fatal `warnings` (and the injected logger). A malformed `package.json` is still treated as "not a workspaces root" — it is not *our* config so we have no standing to fail resolution over it — but it is no longer invisible (no-silent-degradation, P-tier).

- **[LOW] test-coverage — MAX_WALK_UP_DEPTH guard** (`tests/discovery.test.ts`): added a test with a pathological fs and a >64-segment start path where the natural `parent === current` root terminator is never reached, asserting the walk terminates via the depth guard (bounded `existsSync` call count).

- **[LOW] api-contracts — error detail parity**: non-object config errors (in both `loadConfigFile` and `writeUserConfig`) now name the actual parsed JSON type — distinguishing `object` / `array` / `null`, which a bare `typeof` conflates — and carry a `TypeError` cause, matching the context-preserving shape of env-coercion errors. Covered by new tests for `number` and `array` payloads.

- **[LOW] correctness — env-var name regex** (`expand.ts`): tightened from `[^}:-]+` to the POSIX portable charset `[A-Za-z_][A-Za-z0-9_]*`, so `${...}` references containing spaces or punctuation pass through verbatim instead of being treated as expandable. I deliberately did **not** adopt the uppercase-only suggestion (`[A-Z_][A-Z0-9_]*`) verbatim: env var names are case-sensitive and lowercase/mixed-case names are valid, so uppercase-only would silently drop legitimate references. New tests cover both the invalid-name passthrough and lowercase/mixed-case expansion.

No separate changeset: the package is unreleased (0.0.0) and these refinements fold into the existing `config-loader-initial.md` changeset, whose description already covers non-fatal warnings and malformed-file surfacing.

owenjohnson:
Addressed both outstanding maintainer findings on `packages/config-loader/tests/loader.test.ts` (test-only; no production change, no changeset). `make check` green (lint + full test 1341 passed/14 skipped + claudemd-check).

**[MEDIUM] test-coverage — ConfigLogger warning collection.** Strengthened the `warning collection + logger forwarding` suite to assert the full forwarding contract, not just `context.source`:
- Every snapshot warning reaches the injected logger **in order**, carrying the **same message** and the **same `{ source, key }` context** the snapshot recorded (the loader forwards `key` explicitly even when a discovery-sourced warning has none, so the field is always present on the context object).
- Added a **multi-warning** case (two malformed `package.json` ancestors on the walk-up path) proving the fan-out is a loop over the full list — first-nearest-ancestor-first — rather than a first-only forward.
- Added a **no-logger** case proving warnings are still collected on the snapshot when no sink is wired (nothing silently dropped).

**[LOW] test-coverage — MAX_WALK_UP_DEPTH guard.** The 64-iteration guard was already covered against `discoverProjectRoot` directly in `discovery.test.ts`, but the finding names `loader.test.ts` (the public surface). Added a loader-level test that drives the same guard **through `createConfigLoader`**: a pathological fs (every path a non-matching directory) over a 200-segment start path. It asserts no project root is found and that `existsSync` call count is bounded by `64 * 3` walk probes (plus the loader's fixed user-config probe), proving termination came from the depth guard, not the filesystem root.

Note (advisory): the Codex/chatgpt connector findings were reviewed; nothing actionable beyond the above for this round.

owenjohnson (commented):

Verified:
- Read the PR body, changed-file list, local diff against `origin/main`, repo `.agent` standards/patterns, README/API docs, package metadata, and the new loader/discovery/path/flatten/env-expansion source and tests.
- Checked precedence, malformed-file handling, env coercion/expansion, write-back, project-root discovery, and permission claims.

Findings:
1. `packages/config-loader/src/flatten.ts:51` — HIGH — `unflatten()` follows inherited properties while walking dotted keys. A key such as `__proto__.polluted` makes `existing = cursor["__proto__"]` resolve to `Object.prototype`, which `isPlainObject()` accepts, so the next assignment writes onto the global prototype during `write()` serialization. That can be reached from a malicious existing user config or from caller-supplied flat updates. Recommendation: build null-prototype containers and use own-property lookups only, and reject or escape dangerous path segments such as `__proto__`, `prototype`, and `constructor`; add regression tests for both existing-file and update-driven write-back keys.
2. `packages/config-loader/src/loader.ts:398` — HIGH — The write-back path relies on `fs.writeFileSync(..., CONFIG_FILE_MODE)` to enforce `0o600`, but Node only applies `mode` when creating a new file. If `~/.{appName}/config.json` already exists as `0644`/`0666`, `write()` preserves those loose permissions while the README and comments say the user config is written owner-only. This matters because the documented examples include secrets like API keys. Recommendation: after every successful write, explicitly `chmodSync(userConfigPath, CONFIG_FILE_MODE)` and wrap chmod failures as `ConfigError("WRITE_FAILED")`; add a test using the real Node adapter or a fake that models Node’s existing-file mode behavior.

owenjohnson:
Integrated `origin/main` into this branch to clear the sibling-wave merge conflicts (mbot-approved; harness can now merge).

**Conflicted files (1):**
- `CLAUDE.md` — package table. Resolved **keep-both**: kept this PR's new `@centient/config-loader` row AND main's updated `@centient/events` row (structural-logger injection wording). No rows dropped; union preserved. Versions verified in sync via `scripts/sync-claudemd-versions.mjs` (all 7 packages OK, resource count 31).

All other files from main auto-merged cleanly (no content conflicts).

**Verification:** `make check` green — lint passed, tsc 0 errors, vitest 1466 passed / 14 skipped, python tests passed, claudemd-check in sync.

Merge commit pushed to `feat/pkg-config-loader`.

owenjohnson (commented):

Verified:
- Read PR body, current head `9a17c96409b091439c1919b5bc818d76e9b41a6d`, changed-file list, diff, current reviews, and the config-loader source/tests around flattening, write-back, path permissions, and filesystem adapters.

Findings:
1. `packages/config-loader/src/flatten.ts:51` — HIGH — `unflatten()` still follows inherited properties while walking dotted keys. A key such as `__proto__.polluted` makes `existing = cursor["__proto__"]` resolve to `Object.prototype`, which `isPlainObject()` accepts, so the next assignment mutates the global prototype during write-back serialization. This remains reachable from a malicious existing user config or caller-supplied flat updates. Recommendation: build null-prototype containers, use own-property lookups only, and reject or escape dangerous path segments such as `__proto__`, `prototype`, and `constructor`; add regression tests for both existing-file and update-driven write-back keys.
2. `packages/config-loader/src/loader.ts:398` — HIGH — `write()` still relies on `fs.writeFileSync(userConfigPath, serialized, CONFIG_FILE_MODE)` to enforce `0o600`, but the Node adapter delegates to `node:fs.writeFileSync`, where `mode` only applies when the file is created. Existing `0644`/`0666` configs keep their loose permissions even though the README and API contract say write-back stores secrets owner-only. Recommendation: after every successful write, explicitly `chmodSync(userConfigPath, CONFIG_FILE_MODE)` and wrap chmod failures as `ConfigError("WRITE_FAILED")`; add a regression test using the Node adapter or a fake that preserves existing-file mode like Node does.

owenjohnson:
## Round-3 update: merged `main` + addressed all 6 findings (one push)

Pushed `d390202`. This single push (a) resolves the conflict with `main` and (b) addresses every finding from the latest maintainer review (request_id `6839d96a-...`). `make check` (lint + full test + python + claudemd-check) is green; the config-loader suite is 66 tests, the monorepo suite 1752 passed / 14 skipped.

### (a) Merge conflict resolution — KEEP-BOTH

| File | Conflict | Resolution |
|------|----------|------------|
| `CLAUDE.md` | Package table: this branch added the `@centient/config-loader` row; `main` added `@centient/cli-utils` (plus path-security/proc/resilience landed cleanly). | Kept **both** rows. Verified the version column against each `packages/<pkg>/package.json` via `node scripts/sync-claudemd-versions.mjs` (already in sync) and `./scripts/check-claudemd-versions.sh` (all 10 packages + resource count OK). |
| `pnpm-lock.yaml` | Both branches inserted a new importer (`packages/config-loader` vs `packages/cli-utils`) at the same spot. | Kept **both** importer entries, then ran `pnpm install` to canonicalize — "Lockfile is up to date". |

No semantic contradictions; all incoming `main` package additions are purely additive.

### (b) Findings — each at root cause

1. **[HIGH] `paths.ts` / write mode imposes 0o600 silently** — Added a `configFileMode` loader option (default `0o600`), threaded through `write()` → `writeUserConfig`. Callers can now opt into a different mode instead of having owner-only silently reset on existing files; the assert-on-every-write behaviour is documented on the `write()` JSDoc, the option, and the README. Tests: custom mode honoured; default still reasserts/tightens a pre-existing loose file.

2. **[MEDIUM] `loader.ts` `defaultHomeDir()` empty-string** — Added `requireHomeDir()`: an empty/whitespace home (which `os.homedir()` can return, and which would otherwise make `join("", ".app")` a silent cwd-relative path) now throws `ConfigError("INVALID_HOME")`. Applies to both the injected `homeDir` and the `os.homedir()` default. Tests cover empty, whitespace, and valid.

3. **[MEDIUM] `expand.ts` `${VAR}`→"" vs shell** — This is **deliberate, documented seed behaviour** (collapsing to `""` so config templates layer under their baked-in default rather than leaking a literal placeholder); preserving the literal would break that contract. Per the "document the deviation" half of the recommendation, I moved the rationale onto the `expandEnvRefs` function JSDoc (it was only in the module header) and **pinned the contract with two tests** (`${UNSET}`→"", `${UNSET:-}`→"" explicit) so a future "match the shell" change cannot land silently. Behaviour unchanged by design — flagging this finding as **documented/by-design** rather than a code defect, but a real doc+test change accompanies it.

4. **[MEDIUM] `flatten.ts` `unflatten()` silent overwrite** — `unflatten()` now raises `ConfigError("KEY_CONFLICT")` (naming both colliding keys) when dotted keys imply contradictory shapes — both orders: leaf-then-descend (`a.b`, `a.b.c`) and descend-then-leaf (`a.b.c`, `a.b`). Write-back inherits the guard. New `tests/flatten.test.ts` covers both orders, sibling-prefix non-conflicts, and the write-back path.

5. **[LOW] `loader.ts` logger forwarding test coverage** — Added an explicit test that a discovery warning (which carries `source` but no `key`) forwards `key: undefined` as a *present* field in logger context, plus a both-undefined forwarding-shape test. (The existing fan-out test already asserted ordering/parity; these pin the undefined-field contract directly.)

6. **[LOW] `discovery.ts` hardcoded `MAX_WALK_UP_DEPTH`** — Now configurable: `DiscoveryOptions.maxDepth` and the loader-level `projectMaxWalkUpDepth` option, default 64 (exported as `DEFAULT_MAX_WALK_UP_DEPTH`). Non-positive/non-integer values fall back to the default. Tests cover lower cap, raised cap reaching a config 90 levels up that the default would miss, and the fallback.

Notes folded into the existing `.changeset/config-loader-initial.md` (no new changeset, no manual version bump). Zero runtime dependencies retained.

owenjohnson (commented):

Verified:
- Read PR body, current head `d39020277729690529d95875a43527ec7fd94df5`, current review history, PR diff against `origin/main`, and the response delta around `flatten`, `loader`, filesystem adapters, and tests.
- Re-checked the previous prototype-pollution and existing-file permission findings against the current source.

Findings:
1. `packages/config-loader/src/flatten.ts:70` — HIGH — `unflatten()` still reads `cursor[part]` before proving `part` is an own property, so inherited objects remain reachable. For `__proto__.polluted`, `existing` resolves to `Object.prototype`; `isPlainObject()` accepts it because its prototype is `null`, then line 85 moves `cursor` onto `Object.prototype` and the final assignment pollutes the global prototype. The new conflict checks do not reject `__proto__` / `prototype` / `constructor`, and the output containers are still normal `{}` objects. Recommendation: use null-prototype containers, perform own-property lookup before reading a segment, and reject or escape dangerous path segments; add regressions for `unflatten({ "__proto__.polluted": true })` and write-back updates/existing files containing that key.
2. `packages/config-loader/src/loader.ts:409` — HIGH — Existing-file permissions are still not enforced with the Node adapter. The code only passes `fileMode` to `fs.writeFileSync`, and `createNodeFileSystem()` delegates that to `node:fs.writeFileSync(..., { mode })` at `packages/config-loader/src/fs-adapter.ts:24`; Node applies `mode` only when creating a file, so an existing `0644`/`0666` config keeps its loose mode despite the write-back contract. The new fake FS test masks this because `tests/helpers.ts:49` updates the stored mode on every write. Recommendation: call `fs.chmodSync(userConfigPath, fileMode)` after every successful write and wrap failures as `ConfigError("WRITE_FAILED")`; update the fake or add a Node-adapter-backed regression that preserves existing-file mode unless chmod is called.

owenjohnson:
Resolved merge conflicts from sibling wave PRs (integrated `origin/main`), keep-both:

- **CLAUDE.md** (package table): main added the `@centient/dag` row; this PR adds `@centient/config-loader`. Kept both rows. Versions verified in sync via `scripts/sync-claudemd-versions.mjs` (both 0.0.0).
- **pnpm-lock.yaml**: both sides added a workspace-package entry sharing the same trailing devDependencies block (a naive textual keep-both would be wrong). Regenerated the lockfile for the merged 12-project workspace via `pnpm install`; both `packages/config-loader` and `packages/dag` entries are present.

`make check` green: lint + 1792 tests passed (14 skipped) + python-test + claudemd-check. Merge commit `1a75e87` (2 parents) pushed to `feat/pkg-config-loader`.

owenjohnson (commented):

Verified:
- Read PR body, current head `1a75e87c80d37decb7fafbfd5afcd34eebbbe246`, current review history, current PR files, and the latest source for `flatten`, `loader`, filesystem adapters, and tests.
- Confirmed this head did not change the config-loader code paths from the previous reviewed head; the apparent head movement is a rebase/main-history update.

Findings:
1. `packages/config-loader/src/flatten.ts:70` — HIGH — `unflatten()` still reads `cursor[part]` before proving `part` is an own property, so inherited objects remain reachable. For `__proto__.polluted`, `existing` resolves to `Object.prototype`; `isPlainObject()` accepts it because its prototype is `null`, then line 85 moves `cursor` onto `Object.prototype` and the final assignment pollutes the global prototype. The conflict checks still do not reject `__proto__` / `prototype` / `constructor`, and the output containers are normal `{}` objects. Recommendation: use null-prototype containers, perform own-property lookup before reading a segment, and reject or escape dangerous path segments; add regressions for `unflatten({ "__proto__.polluted": true })` and write-back updates/existing files containing that key.
2. `packages/config-loader/src/loader.ts:409` — HIGH — Existing-file permissions are still not enforced with the Node adapter. The code only passes `fileMode` to `fs.writeFileSync`, and `createNodeFileSystem()` delegates that to `node:fs.writeFileSync(..., { mode })` at `packages/config-loader/src/fs-adapter.ts:24`; Node applies `mode` only when creating a file, so an existing `0644`/`0666` config keeps its loose mode despite the write-back contract. The fake FS test still masks this because `tests/helpers.ts:49` updates the stored mode on every write. Recommendation: call `fs.chmodSync(userConfigPath, fileMode)` after every successful write and wrap failures as `ConfigError("WRITE_FAILED")`; update the fake or add a Node-adapter-backed regression that preserves existing-file mode unless chmod is called.

owenjohnson:
Coordinator merge-bot: merging under operator authorization. Verified: maintainer APPROVED on current head `1a75e87` (review commit_id == head SHA), mergeable, base=main.

PR #91, opened 2026-06-13
feat(resilience): add @centient/resilience zero-dep primitives

owenjohnson:
## Summary

New package **`@centient/resilience`** (starts at `0.0.0`, minor-bump changeset) — the first land in the resilience wave. It gives the monorepo a shared, zero-dependency home for the retry/breaker/limiter machinery that `centient`, `mbot`, and `test-kit` currently hand-roll, and unblocks the sdk's configurable-retry upstream ask.

Scope delivered:
- **Circuit breaker** — ported from crucible's pure `CLOSED -> OPEN -> HALF_OPEN -> CLOSED` state machine (timestamps converted to epoch-ms), wrapped in a clock-injected `createCircuitBreaker()` factory with `execute()`/`CircuitOpenError` and out-of-band `onSuccess`/`onFailure`. The pure functions (`checkCircuit`/`recordSuccess`/`recordFailure`) are exported too.
- **Token-bucket rate limiter** — lazy clock-based refill (no background timer), `tryAcquire` returns a `Result`, `acquire` waits via an injectable sleep.
- **Backoff with jitter** — `linear` reproduces the `@centient/sdk` client schedule exactly (base + jitter in `[0, 0.5*base)`) so the sdk can adopt it drop-in; `exponential` with cap also provided. Randomness injected.
- **Cache** — LRU with optional TTL and stale-while-revalidate; `get` returns `{ hit, status: "fresh"|"stale"|"expired", value? }`.
- **Bounded-concurrency pool** — FIFO queue, optional `maxQueue` that sheds load via `PoolRejectedError`, `onIdle()`.

## Design notes

- **Clock + randomness injected throughout.** No `Date.now()`/`Math.random()` in any logic path — both are isolated to single seams (`clock.ts`, `random.ts`). A source-grep test (`primitives.test.ts`) enforces this, mirroring the sdk's existing jitter guard.
- **Factory-function API** (`createCircuitBreaker()`, `createTokenBucket()`, `createBackoff()`, `createCache()`, `createPool()`) and the **`Result`** discriminated-union pattern where failure is an ordinary outcome (rate-limit denial). Throwing is reserved for invalid config.
- Zero runtime dependencies; ESM-only; TS 5.3 strict; mirrors `packages/wal` scaffold conventions.
- README documents every export with examples.

## Test evidence

- `make check` green: `tsc` 0 errors, **vitest 1373 passed / 14 skipped** across the monorepo, claudemd-check OK (`@centient/resilience 0.0.0` row matches `package.json`, resource count 31).
- 77 tests in the new package: circuit-breaker **state-machine tests ported from crucible** (`tests/state/circuit-breaker.test.ts`), property-style backoff bounds (every sampled delay within `[base, base + jitterSpan)`), deterministic rate-limiter/cache via the manual clock, pool concurrency/queue/rejection.

## Followups (NOT in this PR)

- **sdk migration:** adopt `createBackoff({ strategy: "linear" })` in `packages/sdk/src/client.ts` to replace the private `backoffDelay`, and export an `isRetryableError` helper from the sdk.
- **Consumer migrations:** queue PRs in `centient`, `mbot`, `test-kit`, `membrane`, `engram-server`, and `pipeline-sdk` to depend on `@centient/resilience` and drop their hand-rolled breaker/limiter/cache utilities.


owenjohnson:
Addressed all 6 findings from the latest maintainer review (commit 9c53907). `make check` green (lint + 1389 tests + claudemd-check).

- **[MEDIUM] test-coverage — backoff jitterRatio boundary:** Added `jitterRatio = 1.0` (full `[base, base + baseDelay)` envelope), just-below-1.0 (0.999 property sweep under real randomness), and `> 1.0` cases to `backoff.test.ts`. The ratio has no upper bound by design (only negative is rejected), now asserted explicitly.
- **[MEDIUM] test-coverage — circuit-breaker concurrency:** Added a `concurrent execute()` suite using controlled deferreds: several calls admitted while CLOSED resolve independently; N concurrent failures all count and trip OPEN at the threshold (`totalOpens` = 1); a call admitted before the trip still completes while a later success in OPEN state is a graceful no-op.
- **[LOW] error-handling — cache.has() memory leak:** `has()` now evicts a fully-expired entry in passing and counts an `expiration`, mirroring `get()`'s lazy-expiry bookkeeping. A key only ever *probed* (never `get()`) can no longer accumulate dead entries. Root cause, not a periodic sweep. Tests assert size drops to 0 and a fresh entry is untouched.
- **[LOW] api-contracts — rate-limiter silent backward clock:** Surfaced rather than swallowed. The bucket now exposes `clockRegressions()`; a strict backward movement during `refill()` increments it (a stall — `elapsedMs === 0` — does not). This honors the no-silent-degradation principle **without** taking a logger runtime dependency (zero-dep rule): the signal is observable state the caller can alert on. `reset()` clears it. Covered by new tests.
- **[LOW] error-handling — pool next() consistency:** Centralized the slot release into `release()`, which decrements `active` **before** pumping `next()`, inside a `try/finally`. An unexpected throw from `next()` (or the task factory) can no longer strand a queued task with a leaked active count. New test queues a throwing factory behind a busy slot and asserts the follow-up still runs and `active`/`queued` return to 0.
- **[LOW] test-coverage — rate-limiter fractional tokens:** Added fractional-refill-rate (`0.5/s`), partial-interval accrual (`0.5` token), fractional consumption (`tryAcquire(1.5)`), and a 1000-step float-accumulation-stays-capped case.

Notes for the reviewer:
- The `clockRegressions()` counter is a deliberate API addition over the bare "log a warning" suggestion — `@centient/resilience` has **zero runtime dependencies** and no logger seam, so exposing the signal as observable state is the zero-dep-preserving way to satisfy the finding's intent.
- `@centient/resilience` is still unreleased (0.0.0); these behavioral additions fold into the package's existing `minor` changeset, whose description was extended accordingly. No new changeset and no manual version bump.

owenjohnson (commented):

Verified:
- Read the PR body, changed-file list, local diff against `origin/main`, repo `.agent` standards/patterns, README/API docs, package metadata, and the new source/tests for backoff, token bucket, circuit breaker, cache, pool, clock/random/result exports.
- Checked workspace wiring, changeset/package metadata, clock/random injection guards, and the circuit-breaker concurrency tests.

Findings:
1. `packages/resilience/src/circuit-breaker.ts:121` — HIGH — The documented HALF_OPEN contract says a single probe call decides whether to close or re-open, but `checkCircuit()` currently allows every call while `state.state === "half_open"`. In the stateful factory, once one caller transitions OPEN -> HALF_OPEN, concurrent `execute()` calls can all pass the gate before the first probe settles, sending a burst to the recovering upstream and undermining the breaker during outage recovery. Recommendation: add explicit half-open probe ownership/in-flight state in the factory or state machine so only one probe is admitted; reject/short-circuit the rest with `CircuitOpenError`/retry metadata until the probe records success or failure, and add a concurrent regression test for multiple calls arriving in HALF_OPEN before the first settles.

owenjohnson:
Integrated `origin/main` into this branch to clear the sibling-wave merge conflicts. Resolved KEEP-BOTH:

- **CLAUDE.md** (package table): main added the `@centient/proc` row; this PR adds `@centient/resilience`. Kept both rows (ordered proc → resilience). Versions verified against each `package.json` via `scripts/sync-claudemd-versions.mjs` — all in sync.
- **pnpm-lock.yaml**: kept both `packages/proc` and `packages/resilience` workspace entries; lockfile regenerated with `pnpm install` to stay authoritative.

`make check` green: lint passed, tsc 0 errors, vitest 1514 passed / 14 skipped, claudemd-check OK (resource count 31). No PR logic changed — merge integration only.

🤖 merge commit 05d6f23

owenjohnson:
Coordinator merge-bot: merging under operator authorization. Verified: maintainer APPROVED on current head `05d6f23` (review commit_id == head SHA), mergeable, base=main.

PR #95, opened 2026-06-13
docs(tradeoffs): five deferred doc tradeoffs + sdk node-20 floor (Initiative 2)

owenjohnson:
## Summary

Initiative 2 of `docs/plans/2026-06-12-next-stage.md`: batch-merge the five "documented tradeoff" items from the phase-6 audit (`docs/hardening/2026-06-10-dimensions.md`, Deferred lines 47-57) into merged README text, plus one metadata-only `package.json` field. These were tradeoffs accepted *conditional on being documented loudly*; undocumented they read as silent degradation (P2). Text-only except the engines field; no behavior change.

## What changed

- **wal README — "Crash window" subsection** (under dead-lettering): confirm-before-append (`packages/wal/src/replay.ts:210-254`) means a crash between confirming the original entry and appending the `dead_letter` record silently loses the dead-letter trail for that one entry but never double-executes. Documents the explicit choice of **no-double-execution over no-loss** and bounds the window to the retry-cap boundary.
- **sdk README — "Server availability"**: thin client, **no graceful degradation** (no cache/offline queue/fallback); when engram-server is down every call rejects. Built-in retry is for transient faults only; any retry/backoff beyond it is the caller's responsibility.
- **sdk README — "Runtime requirements"**: per-request timeout and connection-establishment abort are built on global `fetch` + `AbortController`/`AbortSignal` (`client.ts`); states the supported floor (**Node >= 20.0.0**) explicitly.
- **events README — `onEvent()` must-not-block contract**: tee'd callbacks fire synchronously inside `emit()` and are not awaited; documents that the configured `BackpressurePolicy` applies **only** to `subscribe()` AsyncIterable buffers, never to the `tee()` callback path (which has no buffer) — answering the spec's "which policy applies to callback fan-out" with "none; use subscribe() for bounded delivery."
- **sdk `package.json` — `engines.node` `>=18.0.0` → `>=20.0.0`**: enforces the fetch/AbortSignal floor at the strongest level and aligns with events/wal and the repo CLAUDE.md. Metadata-only; shipped as a **patch changeset** for `@centient/sdk` with a CHANGELOG-worthy body (`.changeset/sdk-engines-node20-floor.md`).
- **dimensions doc annotations**: the four implemented Deferred items annotated "closed by PR #<this>"; the fifth (JSONL remainder buffer 1 MB → 100 KB) annotated "re-deferred, see next-stage spec" per the spec's Deferred Again section.

## Design notes

- All README additions are self-contained new sections (`## Server availability`, `## Runtime requirements`, `#### onEvent() contract`, `### Crash window`) to minimize conflicts.
- **Concurrency note:** initiative-6 (logger injection in events/wal) and initiative-7 (sdk event subscription) are editing other sections of these same three READMEs in parallel. My additions are isolated sections to keep merge conflicts trivial; a textual conflict at most, no semantic overlap.
- No new exported symbols; the engines change is the only non-doc edit.

## Test evidence

- Acceptance greps all hit: `crash window` (wal), `graceful degradation` / `server is down` (sdk), `must not block` (events), `node` floor statement (sdk).
- `node -e "...engines.node"` prints `>=20.0.0`.
- `pnpm build` — 5/5 packages green.
- `pnpm test` — full suite green.
- `make check` — green: `tsc 0 errors`, `vitest 1296 passed / 14 skipped`, claudemd-check "All package versions in CLAUDE.md match."

## Followups

- None blocking. The dimensions-doc annotation is amended into this same PR after open (see final commit).


owenjohnson:
Addressed all three findings from the latest maintainer review (`request_id: 99bb7860...`). Pushed as d4f6224.

**1. [LOW] api-contracts — sdk changeset bump level (FIXED, root cause)**
The last published `@centient/sdk` declares `engines.node: ">=18.0.0"`, so raising the floor to `>=20.0.0` *is* consumer-visible: Node 18 installers get an `EBADENGINE` warning and a hard failure under `--engine-strict`. A `patch` understated that. Reclassified the changeset from **patch → minor** and documented the consumer impact in the changeset body. (Not `major`: there is no API/type/runtime break — only a tightening of the declared runtime floor — so a minor bump matches the actual contract change.)

**2. [LOW] error-handling — events callbacks-must-not-block has no runtime guard (rationale + doc, finding partially misapplied)**
Adding runtime detection (timing every callback / timeout wrappers) would itself be a behavior + per-event-overhead change to the synchronous `tee()` fan-out — the exact path whose stated value is being unbuffered, allocation-free, and immediate — and this PR is the *documentation* of that accepted tradeoff, not a redesign of it. A wall-clock threshold also cannot distinguish a legitimately slow consumer from a buggy one without false positives. So rather than instrument the hot path, I documented **why** the contract is enforced by convention and pointed callers to the existing **structural** escape hatch: if you cannot guarantee a non-blocking callback, use `subscribe()` (AsyncIterable + `BackpressurePolicy`) instead of `tee()`. The smallest honest change that satisfies the finding's spirit (no silent trap) without a perf/behavior regression.

**3. [LOW] test-coverage — wal crash window has no test (FIXED, root cause)**
Added a deterministic test (`dead-letter crash window` in `packages/wal/tests/wal.test.ts`) that pins the documented **confirm-before-append** invariant. Rather than fake an OS crash, it injects the failure at the exact seam the README names — a hoisted `node:fs/promises` mock fails the dead-letter append's `open(walPath, "a")` (the only append-mode open in that flow) *after* `confirmEntry` has committed. It then asserts the three documented guarantees: the original entry stays `confirmed: true` (no re-replay / no double execution), no `dead_letter` record is written (the accepted silent loss), and a subsequent replay does not re-run the operation. The default mock is a pass-through, so the other 63 WAL tests are unaffected.

Gate: `make check` green — tsc 0 errors, vitest **1297 passed / 14 skipped** (was 1296; +1 is the new crash-window test), claudemd-check "All package versions in CLAUDE.md match."

owenjohnson:
Coordinator merge-bot: merging under operator authorization. Verified: maintainer APPROVED on current head `d4f6224` (review commit_id == head SHA), mergeable, base=main.

PR #96, opened 2026-06-13
feat(proc): new @centient/proc hardened subprocess runner

owenjohnson:
## Summary

Adds **`@centient/proc`** — a binary-agnostic hardened wrapper over `node:child_process` `spawn`, extracted from the generic subprocess machinery in membrane's `DockerCLI` (lines ~81-245, ~332-404). Nothing docker-specific was carried over; verified by grep for docker/membrane identifiers across src/tests/README/package.json (clean).

The package solves, once, the concerns every robust process call needs: timeouts, kill escalation, output buffer caps, cancellation, stdin streaming, and a single typed error.

## What it does

`runProcess(command, options?): Promise<ProcResult>`:
- **Wall-clock timeouts** with **SIGTERM-then-SIGKILL** escalation (configurable `killGraceMs`); never relies on the child cooperating.
- **Per-stream stdout/stderr byte caps** — overflow kills the process rather than growing memory unbounded.
- **`AbortSignal`** cancellation, both pre-spawn and mid-run.
- **stdin streaming** with EPIPE tolerance.
- **One typed `ProcError`** whose `kind` discriminates `spawn-failure` / `non-zero-exit` / `timeout` / `signal` / `buffer-overflow` / `aborted`. No message parsing; `isProcError()` narrows.
- **Settle-once promise** — every terminal path (close/error/timeout/abort/overflow) funnels through a single gate, so a double-resolve/reject is structurally impossible.

## Design notes

- **Zero external runtime dependencies**; `node:child_process` only. ESM-only, TS 5.3 strict, NodeNext. Scaffold mirrors `packages/wal`.
- **Observable architecture:** the `Clock` (timeout/kill timers) and the `spawn` impl are injectable. No ambient `Date.now()`/`Math.random()` in core paths — timeout and kill escalation are driven through the injected clock, so they are tested deterministically with no real sleeps.
- **No shell interpretation** of `command`/`args` — callers aren't exposed to shell injection through this package.
- New package at `version: 0.0.0` with a `.changeset/proc-initial.md` declaring a **minor** bump. CLAUDE.md table row added (version column = `0.0.0`, claudemd-check green).

## Test evidence

15 tests, **under 1s**:
- Real `node -e` commands: success exit codes, non-zero-exit (output preserved), ENOENT spawn-failure, stdin round-trip, buffer-mode raw output, real short-timeout kill, AbortSignal kill, pre-aborted reject, stdout buffer-overflow kill.
- Fake child + injected clock: deterministic SIGTERM→SIGKILL escalation, timer cleared on clean exit (no kill), external-signal error mapping.
- **Settle-once invariants:** timeout-then-late-close, overflow-then-close, error-after-success — each asserts exactly one settle.

`make check` is green from repo root: tsc 0 errors, vitest **1311 passed / 14 skipped**, claudemd-check `OK @centient/proc (0.0.0)` + resource count 31.

## Followups (consumer-migration PRs to queue)

- **membrane `DockerCLI` re-base** onto `@centient/proc` — behavioral no-op asserted by its existing suite (the docker-specific error mapping stays in membrane; only the generic spawn/timeout/buffer/settle machinery moves under the runner).
- **test-kit container manager** — adopt `runProcess` for container lifecycle calls.
- **soma-test-harness runner** — replace ad-hoc spawn wrappers.
- **daemon spawn helpers** — use the runner for managed daemon launches.


owenjohnson:
## Independent review fix

Verify-don't-trust adversarial pass over `@centient/proc` (FOCUS: promise double-settle, SIGTERM→SIGKILL race, buffer-cap overflow, AbortSignal, timer/handle leaks).

The settle-once gate, double-settle guard, buffer-cap overflow path, and AbortSignal handling (pre-spawn + post-spawn) all held up — no bug there.

### Bug found + fixed: SIGKILL-escalation timer leaks after the child exits

**Location:** `packages/proc/src/run.ts` — `escalateKill()` / `close` + `error` handlers.

The kill-grace timer is armed *after* `settle()` runs (so `clearTimers()` deliberately cannot reach it — that guarantees a child ignoring SIGTERM is still force-killed). But once the child actually exits, the `close`/`error` handlers call `settle()`, which is a no-op after the first settle and never cancels the pending `killTimer`. Result: on every timeout / abort / buffer-overflow where the child obeys SIGTERM and exits, a `killGraceMs` timer (default 5s) **dangles for the full grace window** — holding the event loop open and later firing a redundant `child.kill("SIGKILL")` at an already-dead, possibly PID-recycled process.

**Fix:** added `cancelKillTimer()` and call it from the `close` and `error` handlers (which fire only when the child is genuinely gone). The escalation path is preserved: if the child stays alive past the grace window, SIGKILL still fires.

**Proving tests** (added to `tests/run.test.ts`, fail-before / pass-after, verified by stashing the fix):
- "clears the SIGKILL-escalation timer once the child closes after a timeout" — asserts `clock.pending === 0` and no redundant SIGKILL after the child closes following a timeout. **Failed before** (pending timer = 1), passes after.
- "clears the SIGKILL-escalation timer once the child closes after an abort" — same, via AbortSignal. **Failed before**, passes after.
- "still escalates to SIGKILL when the child ignores SIGTERM" — guards the fix from over-cancelling; passes in both states.

`make check` green (1314 passed, 14 skipped; tsc 0 errors; CLAUDE.md table in sync). No new changeset — folds into the existing `.changeset/proc-initial.md` for the unreleased 0.0.0 package.

owenjohnson:
Addressed all 8 findings from the latest maintainer review (commit 43b98e2). Summary:

**[HIGH] security — command/args validation (spirit, not literal):** There is no shell-injection surface to sanitise: `runProcess` never sets `shell: true`, so `command`/`args` go straight to `spawn` as a program image + argument vector — nothing is tokenised, glob-expanded, or interpolated. `"foo; rm -rf /"` becomes a single bogus program name (→ `spawn-failure`), never two commands. A whitelist/metacharacter filter would only break the deliberately binary-agnostic contract (callers legitimately pass arbitrary absolute paths and argument bytes). I did fix the one genuine foot-gun — an empty / non-string `command` is now rejected eagerly before touching the OS — and documented the no-shell stance in `run.ts`, `types.ts`, and the README. Added a test proving a metacharacter-laden command + args are handed to spawn verbatim.

**[HIGH] error-handling — stdin handler swallowed all errors:** Now distinguishes expected pipe teardown (`EPIPE` / `ERR_STREAM_DESTROYED`, still ignored — the child just stopped reading) from genuine I/O failures (e.g. `ENOSPC`). An unexpected stdin error on an otherwise-clean (exit 0) run surfaces as a new `stdin-error` `ProcError` kind rather than a misleading success (repo principle: no silent degradation). A non-zero exit still wins as the headline failure. Three new tests.

**[MEDIUM] error-handling — killGraceMs: 0 sent SIGTERM+SIGKILL:** `escalateKill` now sends SIGKILL only when grace is 0 (the same-tick SIGTERM was pure noise). Documented on the option; test asserts `["SIGKILL"]` with no escalation timer armed.

**[MEDIUM] api-contracts — encoding only utf8/buffer:** `encoding` now accepts any Node `BufferEncoding` (`hex`, `base64`, `latin1`, `utf16le`, …) plus `"buffer"`, via a new exported `ProcEncoding` type. Decode is applied once to the concatenated stream (so multi-byte sequences split across chunks aren't corrupted). Tests for hex/base64/latin1 and the split-multibyte case.

**[MEDIUM] test-coverage — edge cases:** Added (1) simultaneous stdout+stderr overflow in one tick (settles once), (2) 256 KiB input round-trip (> default 64 KiB pipe buffer), (3) env inherited-when-undefined vs restricted-when-provided, (4) special-character command/args passed verbatim.

**[LOW] error-handling — overflow message lacked accumulated bytes:** `buffer-overflow` errors now carry `actualBytes` and the message reads `…exceeded N byte cap (accumulated M bytes)…`. Test asserts both fields.

**[LOW] api-contracts — no global default config (rationale):** Declined a `.configure()`/global-mutable-defaults mechanism — `runProcess` is a pure function with no ambient state, which is what keeps it deterministic and trivially testable. Defaults are already per-call overridable; callers wanting an app-wide default wrap `runProcess` once. Documented this on the `DEFAULT_*` exports.

**[LOW] test-coverage — pre-aborted should prove no spawn:** Added a counting spawn impl; new tests assert spawn is never invoked for a pre-aborted signal, an empty command, or a non-string command.

`make check` is green: tsc 0 errors, 1331 tests pass (35 in proc), claudemd-check OK. No new runtime deps; ESM-only; lockfile unchanged.

owenjohnson:
## Remediation — round addressing the 2 persistent findings (commit 23972de)

### [HIGH] security — `packages/proc/src/run.ts` command/args validation

Addressed in spirit, not by adding a metacharacter filter or executable whitelist — those would be band-aids that break the package's load-bearing contract. The runner never sets `shell: true`; `command`/`args` go straight to `spawn` as a program image + argument vector. `"foo; rm -rf /"` is a single bogus program name (→ `spawn-failure`), never two commands, so there is no *shell*-metacharacter injection surface to sanitise. A whitelist would break the deliberately binary-agnostic contract (callers legitimately pass arbitrary absolute paths and arbitrary argument bytes — spaces, UTF-8, binary).

What I did change closes the one genuinely **shell-agnostic** injection vector the prior rounds had not: a **NUL byte (`\0`)** in the command or any arg. A NUL silently truncates the C-string the underlying syscall sees, so `"safe\0; rm -rf /"` would drop everything after the NUL rather than pass it verbatim — violating the "args go through untouched" guarantee. Node otherwise rejects this deep in libuv with an opaque `ERR_INVALID_ARG_VALUE`; the runner now rejects it eagerly with an attributable, typed `spawn-failure` (the arg case names the offending index). Every other byte is still forwarded unchanged.

- Validation added in `run.ts` after the existing empty/non-string guard, with a doc comment explaining why this is the *only* honest input guard here.
- Documented in the README no-shell section and the changeset.
- Tests: rejects a NUL in the command and rejects a NUL in an arg, both asserting `spawn` is never invoked (counting spawn impl, `calls.length === 0`).

### [MEDIUM] test-coverage — `packages/proc/tests/run.test.ts` edge cases

Three of the four called-out cases already had coverage on this branch: simultaneous stdout+stderr overflow in one tick (`settles once when stdout AND stderr both overflow`), input larger than the OS pipe buffer (256 KiB round-trip), and command/args with special characters (`passes a command with shell metacharacters verbatim`). The genuine gap was the **env** case, now closed:

- `forwards env verbatim to spawn (undefined stays undefined, never clobbered)` — inspects the options object the runner hands to spawn and asserts an omitted `env` reaches spawn as `undefined`, so spawn applies its own inherit-parent default rather than the runner imposing `{}` or `process.env`.
- `forwards an explicitly-empty env object verbatim to spawn` — an explicit `{}` (the caller's deliberate no-inherited-env choice) is passed through unchanged.

### Gate

`make check` green: tsc 0 errors, 1335 tests pass (39 in `@centient/proc`), claudemd-check OK. No new runtime deps; ESM-only; lockfile unchanged. Folded into the existing `.changeset/proc-initial.md` (unreleased 0.0.0 package).

owenjohnson (commented):

Verified:
- Read the PR body, changed-file list, local diff against `origin/main`, `packages/proc/src/run.ts`, public types/errors, README/API docs, package metadata, lockfile entry, and the runner tests.
- Checked timeout/abort/overflow settle paths, stdin handling, no-shell/NUL guards, env forwarding, and package exports.

Findings:
1. `packages/proc/src/run.ts:289` — HIGH — after the first buffer overflow rejects the promise, later `data` events still call `CappedBuffer.push()` and keep storing output while the child is being killed. That means a process that ignores SIGTERM and keeps writing during `killGraceMs` can continue growing memory despite the advertised per-stream cap. Those later overflows also call `escalateKill()` again, potentially arming multiple SIGKILL timers while only the last handle is tracked/cancelled. Recommendation: make overflow termination one-shot (for example have `settle` return whether it won), stop/destroy the stdio readers or ignore further data after settlement/overflow, and add a regression with `killGraceMs > 0` that emits multiple over-cap chunks before close.
2. `packages/proc/src/run.ts:203` — MEDIUM — synchronous exceptions from `spawnImpl()` bypass the promised typed-error contract. Node can throw before returning a child for invalid runtime inputs/options (and injected spawners can throw too); because the call is not wrapped, callers receive a raw `TypeError`/custom error instead of `ProcError("spawn-failure")`. Recommendation: wrap the `spawnImpl(command, args, ...)` call in `try/catch` and reject with a `ProcError` carrying the original cause, with a test using a throwing spawn implementation.

owenjohnson:
Coordinator merge-bot: merging under operator authorization. Verified: maintainer APPROVED on current head `23972de` (review commit_id == head SHA), mergeable, base=main.

PR #87, opened 2026-06-13
chore(monorepo): close the release-gate gap — mechanical CLAUDE.md sync + hardened publish (Initiative 1)

owenjohnson:
## Summary

Next-stage **Initiative 1, steps 2–4** (`docs/plans/2026-06-12-next-stage.md`; step 1 landed as #82). Closes the structural gap that made every release leave main red on `claudemd-check`.

### Root cause (step 4 — recorded in `docs/hardening/STATE.md` phase 7)

Reconstructed from git reflog + npm registry timestamps: the 2.0.0/0.7.0 publish was **one uninterrupted `make publish`** (sdk 2.0.0 hit npm 14 s after the `chore: version packages` commit). No operator bypass, no direct `changeset publish` — the T5 stamp gate worked exactly as built. The flaw was sequencing: every gate validated the **pre-bump** tree; `changeset version` then created the CLAUDE.md drift and nothing re-validated before publish+push. RELEASING.md even documented the table sync as a manual post-release step.

### Fix (steps 2–3)

- **`scripts/sync-claudemd-versions.mjs`** — rewrites the CLAUDE.md package-table version column from `packages/*/package.json`; fails loudly on a missing row (description column stays human-curated). Chained as `pnpm run version-packages` (`changeset version && node scripts/sync-claudemd-versions.mjs`).
- **`make publish` hardening** —
  - refuses on a dirty tree or `HEAD != origin/main` (fresh fetch) **before any mutation**;
  - uses `pnpm run version-packages` so the bump commit carries the CLAUDE.md sync;
  - **re-runs the full `make check` live against the post-bump tree** in the same invocation, immediately before `changeset publish` — a stale stamp cannot satisfy it;
  - `DRY_RUN=1` exercises all gates and stops (exit 0) before version bump / npm.
- **Docs** — RELEASING.md gate list rewritten (6 gates); `make publish` declared the only publish path and bare `pnpm changeset version` / direct `pnpm changeset publish` forbidden in `.agent/procedures/commits.md`; CLAUDE.md common-commands updated.
- **Ledger** — STATE.md phase-7 root-cause note; T7 + #67 marked resolved (were stale-open); BACKLOG.md T5 annotated (spec item 3 held as written but was insufficient — pre-bump only).

### Verification

[...]

PR #105, opened 2026-06-13
feat(sdk-python): 0.34.0 parity — Phase A (truthful de-scoping) + Phase B (core)

owenjohnson:
## Summary

Initiative 3 of `docs/plans/2026-06-12-next-stage.md` — bring `sdk-python` back toward the 0.34.0 server contract, and (most urgently, per P11) stop the package from claiming parity it lost when `@centient/sdk` went 1.7.1 → 2.0.0.

**Phase A — stop lying.** The stale `"Full API parity"` claim is replaced by an explicit **support matrix** (resource × supported?) in both README and CHANGELOG. The historical 1.0.0 CHANGELOG claim now carries a correction note. No unqualified full-api-parity claim survives a case-insensitive grep (remaining hits are the negation in the README, the two correction notes, and the historical line that the correction note directly precedes).

**Phase B — 0.34.0 core.**
- `expected_version` + `skip_embedding` on crystal create/update params, mirroring TS semantics including the server-floor caveats. An explicit `skip_embedding=False` is sent on the wire (explicit caller signal); `None` omits it — matching TS.
- `CrystalVersionConflictError` mapped from HTTP 409 `OPERATION_VERSION_CONFLICT` (both flat `{code,message}` and nested `{error:{...}}` shapes), carrying `current_version`, routed before the generic 409 `SessionExistsError`.
- `MaintenanceResource` / `SyncMaintenanceResource` (`client.maintenance`) with `vacuum()` / `tombstone_cleanup()` / `changelog_compact()` handling **bare** (non-enveloped) bodies; a wrapped `{ data }` body fails loudly with `INTERNAL_ERROR` (mirrors the loud-shape-guard in `packages/sdk/src/resources/maintenance.ts`).
- `MIN_SERVER_VERSION = "0.31.0"` + `check_server_compatibility()` mirroring `packages/sdk/src/client.ts:204` — calls `/health`, compares against the floor, fails closed on an unknown/below-floor version.

## Design notes

- **Wire serialization is automatic and verified.** The crystal params use `alias_generator=to_camel` + `exclude_none=True`, so `expected_version`/`skip_embedding` serialize to `expectedVersion`/`skipEmbedding`. Tests assert the exact sent body.
- **No `Math.random`/clock in core paths.** `_is_version_gte` is pure; compatibility reads the server's reported version.
- **Bare-body guard** rejects a `{ data }`-wrapped maintenance body explicitly, so an un-migrated/regressed server fails loudly rather than returning a model with missing fields.
- **pyproject stays `1.0.x`.** Phase C (sync resource + 6 remaining resources + 2.0.0-aligned version bump) is deferred per the spec. `sdk-python` has no `package.json`, so it is invisible to Changesets/pnpm workspaces — no JS changeset applies.

## Gate wiring (spec step 4)

- `scripts/run-python-tests.sh` auto-creates a **gitignored** `packages/sdk-python/.venv`, editable-installs the package + dev extras, and runs the **core** unit suite (`-m "not integration"`, excluding the three framework-example adapter tests which exercise optional `examples/` code and are red on a clean checkout regardless of this change).
- `make python-test` runs that script; it is wired into `make check`.
- **Merge-with-#87 caveat:** the Makefile on `origin/main` differs from the in-flight PR #87 that restructures the `publish` target. I touched **only** the `.PHONY`, the `check` dependency list, and the new `python-test` target lines — nothing in `publish` — to keep that merge trivial.

## Test evidence

`make check` green from the repo root:
- lint: **0 errors**
- vitest (TS): **1296 passed**, 14 skipped
- pytest (sdk-python): **538 passed**, 10 deselected (513 baseline + 25 new)
- claudemd-check: all package versions match, resource count 31

New tests (mocked transport, mirroring the TS resource-test discipline — assert path/method/body + parse enveloped AND bare shapes): `tests/test_maintenance.py`, `tests/test_compatibility.py`, `tests/test_crystal_cas.py`.

## Followups

- **Phase C** (next-stage spec Initiative 3, step 3): `SyncResource` (NDJSON push/pull, peers, conflicts) against the `{success,data}` envelopes, plus `agents`, `ambient_context`, `facts`, `gc`, `memory_spaces`, `users`; then the 2.0.0-aligned version bump via the release procedure. Reference: `packages/sdk/src/resources/sync.ts`.
- The three framework-example integration tests (`langchain`/`crewai`/`autogen`) are pre-existing red on a clean checkout (Mock setup, not SDK code). Excluded from the gate; worth a separate fix.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ SUMMARY := . $(TOOLKIT)/common.sh && . $(TOOLKIT)/summary.sh
 
 .DEFAULT_GOAL := help
 
-.PHONY: help install ensure-deps build lint test check clean publish claudemd-check
+.PHONY: help install ensure-deps build lint test python-test check clean publish claudemd-check
 
 help: ## Show available targets
 	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-15s\033[0m %s\n", $$1, $$2}'
@@ -36,7 +36,10 @@ test: ensure-deps ## Run tests
 CHECK_STAMP := .logs/.check-stamp
 TREE_FINGERPRINT := { git rev-parse HEAD; git status --porcelain; git diff HEAD; } | git hash-object --stdin
 
-check: lint test claudemd-check ## Run full CI gate (lint + test + docs drift)
+python-test: ## Run the sdk-python core unit suite (mocked transport)
+	@$(SUMMARY) && run_summarized generic "./scripts/run-python-tests.sh" .logs/python-test.log
+
+check: lint test python-test claudemd-check ## Run full CI gate (lint + test + python + docs drift)
 	@mkdir -p .logs && $(TREE_FINGERPRINT) > $(CHECK_STAMP)
 
 clean: ## Remove build artifacts
@@ -82,3 +85,8 @@ claudemd-check: ## Check CLAUDE.md package table matches actual versions
 # 2026-06-11  Hard-gate publish on check via tree fingerprint stamp;
 #             explicitly decline npm provenance (NPM_CONFIG_PROVENANCE=false)
 #             for local publishes — unsupported without a CI OIDC provider
+# 2026-06-12  Add python-test (sdk-python pytest via scripts/run-python-tests.sh)
+#             and wire it into `check` (next-stage Initiative 3, Phase B). Only
+#             the .PHONY + check + new python-test lines were touched, to keep
+#             the merge with the in-flight publish-target restructure (PR #87)
+#             trivial.
--- a/packages/sdk-python/.gitignore
+++ b/packages/sdk-python/.gitignore
@@ -1,0 +1,14 @@
+# Local development virtualenv (see README "Development")
+.venv/
+venv/
+
+# Python build / cache artifacts
+__pycache__/
+*.py[cod]
+*.egg-info/
+build/
+dist/
+.pytest_cache/
+.coverage
+htmlcov/
+.mypy_cache/
--- a/packages/sdk-python/CHANGELOG.md
+++ b/packages/sdk-python/CHANGELOG.md
@@ -1,5 +1,54 @@
 # Changelog
 
+## Unreleased
+
+### Parity correction — TypeScript SDK 2.0.0
+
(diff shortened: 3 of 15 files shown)


owenjohnson:
Addressed all 6 findings from the latest maintainer review (commit c82dffc). `make check` green (lint + vitest 1296 + python 546 + claudemd-check).

| # | Finding | Fix |
|---|---------|-----|
| 1 | MEDIUM error-handling — `_require_bare_object` loses the original response | Drift errors now embed a length-capped (200-char), `repr`-safe excerpt of the actual body via `_truncate_body()`. `repr` neutralizes control chars; the cap prevents dumping an unbounded/sensitive payload. New tests pin both the excerpt and the truncation. |
| 2 | MEDIUM api-contracts — vacuum builds query via string concat | `vacuum()` now routes through a `_vacuum_query_params()` helper that drops `None`/falsey values and hands the mapping to httpx's `params` plumbing (central, consistent encoding; future params slot in here). Path stays bare. Tests updated + a new `test_vacuum_default_sends_no_query_params`. |
| 3 | LOW test-coverage — no retry test for transient errors on CAS update | Added `test_transient_5xx_during_cas_update_is_retried` (one 503 then success; asserts `expectedVersion` survives the retry) and `test_transient_5xx_exhausts_retries_then_raises`. |
| 4 | LOW type-safety — `_extract_current_version` allows negative versions | Now requires `value >= 0` and excludes `bool` (an `int` subclass) so `-1`/`True` return `None` rather than a bogus CAS hint. Two new tests. |
| 5 | LOW test-coverage — no pre-release/build-metadata version tests | Added `test_prerelease_and_build_metadata_fail_closed` covering `0.31.0-alpha`, `0.34.0-rc.1`, `0.31.0+build123`, `0.31.x`. These remain fail-closed by design (matches the TS SDK, which only parses the numeric triple) — see finding 6 for the new diagnosability. |
| 6 | LOW error-handling — `_is_version_gte` silently returns [] on unparseable segments | `parse()` now emits a `logger.warning` naming the bad segment before failing closed, so the format issue is diagnosable in production. New `test_unparseable_segment_logs_warning`. |

No changeset: changes are confined to the unversioned `sdk-python` package (table row `-`, not in the JS changesets flow), consistent with this PR's prior commit.

owenjohnson:
Coordinator merge-bot: merging under operator authorization. Verified: maintainer APPROVED on current head `c82dffc` (review commit_id == head SHA), mergeable, base=main.

PR #108, opened 2026-06-13
feat(events,wal): unify logger injection via structural-logger pattern

owenjohnson:
## Summary

Initiative 6 of `docs/plans/2026-06-12-next-stage.md` — `@centient/events` and `@centient/wal` adopt the SDK's injectable structural-logger pattern (`packages/sdk/src/logging.ts`). Each package now defines a minimal structural logger interface and accepts it as an optional option on every entry point, so a consumer can route package-internal logging to their own logger instance. Default behavior is identical to today: `@centient/logger` remains the default, constructed lazily behind a single injection point.

## What changed

**events** (`EventsLogger`: `debug`/`info`/`warn`/`error`, context-first or message-only):
- `createEventStream({ logger })`, `fromJsonl(path, { logger })`, `createJsonlSubscriber(path, { logger })`.
- Default component names preserved byte-for-byte (`centient:events`, `events:jsonl`, `events:replay`) via a per-component lazy cache.

**wal** (`WalLogger`, same shape):
- Options object: `appendEntry`, `replayUnconfirmed`, `replayAndCompact`.
- Optional trailing arg: `readEntries`, `confirmEntry`, `getUnconfirmedEntries`, `compactWal`, `cleanupOrphanedTempFiles`.
- `clearRetryCounts()` untouched (per spec).
- Replay forwards its logger to the read/confirm/compact/append calls it drives, so all WAL-internal logging during a replay routes to one logger.

## Design notes

- **Single default-construction site per package.** `grep -rn "createComponentLogger" packages/events/src packages/wal/src` shows exactly one construction call per package, inside `logging.ts` behind `resolveLogger()`. Every other module resolves through that injection point.
- **Lazy default.** `@centient/logger` stays a runtime dependency (it IS the default impl, per the spec's decision rule), but the component logger is built only on first use — importing the package or injecting a custom logger never constructs it.
- **Structural, not nominal.** The interface is shape-compatible with `@centient/logger`'s `Logger`; a real `Logger` satisfies it directly, and any object with the four methods works. The `info` level is included because the events/wal internals genuinely use it (the spec's `{debug,warn,error}` sketch was illustrative).
- **Public API additions only.** All additions are optional fields / optional trailing args — no signature breaks. d.ts reviewed (`EventsLogger`/`WalLogger`/`JsonlSubscriberOptions` exported from each index; `resolveLogger` is `@internal` and not re-exported from the package entry).

## Test evidence

- New `packages/events/tests/logger-injection.test.ts` and `packages/wal/tests/logger-injection.test.ts`: inject a capture logger (`createTestLogger` from `@centient/logger` — eating our own testing utilities) and assert internal warn-path messages arrive (closed-stream emit / JSONL serialization failure for events; malformed-line read + replay retry-pending for wal). Each also has a default-path regression test asserting unchanged delivery/round-trip behavior when no logger is passed, plus a "any structural logger works" case.
- `make check` fully green: lint (tsc 0 errors) + full suite (1305 passed, 14 skipped) + claudemd-check (all versions match, resource count 31).

## Changesets

Minor for both `@centient/events` and `@centient/wal`. No version bumped manually; no new runtime dependency (testing utils come from the already-present `@centient/logger`), so `pnpm-lock.yaml` is unchanged.

## Followups / overlap flag

- **README overlap:** the initiative-2 and wal-atomic-exports agents are concurrently editing other sections of these READMEs. My additions are a self-contained "Logger injection" section placed immediately before the License section in each README to minimize merge conflicts — flagging for the merge order.
- CLAUDE.md `@centient/events` / `@centient/wal` rows updated to mention logger injection (version columns unchanged at 0.2.3 / 0.3.3, matching package.json).


owenjohnson:
## Independent review fix

**Bug (default-path log regression — contract/parity):** `createEventStream().jsonl()` forwarded the stream's **resolved** default logger to `createJsonlSubscriber(filePath, { logger })`. In the no-injection path that resolved logger carries the `events` component, which then overrode the JSONL subscriber's own `events:jsonl` component. Result: default-path JSONL diagnostics (write/serialization/flush errors) printed under `[events]` instead of the historical `[events:jsonl]`.

This violates the stated invariant of this PR — "default-path log output is byte-for-byte what it was before injection." Before this PR, `stream.jsonl()` called `createJsonlSubscriber(filePath)` with no logger, so JSONL diagnostics always carried `component: "events:jsonl"`.

**Location:** `packages/events/src/stream.ts` — `createEventStream()` (`const logger = resolveLogger(opts?.logger)`) and the `jsonl()` method (`createJsonlSubscriber<T>(filePath, { logger })`).

**Fix:** Keep the raw injected logger (`opts?.logger`, possibly `undefined`) separate from the resolved one. Stream-internal paths keep using the resolved `logger`; the JSONL subscriber receives only the *injected* logger. When none is injected, the subscriber resolves its own `events:jsonl` default exactly as before. The injected path is unchanged — diagnostics still route to the caller's logger.

**Proving test:** `packages/events/tests/logger-injection.test.ts` → "emits JSONL diagnostics under the events:jsonl component (not events)". It captures the default logger's `console.error` output, drives the default-path `stream.jsonl()` with a non-serializable (`BigInt`) event, and asserts the diagnostic line contains `[events:jsonl]` and not `[events] `. Fails on the pre-fix code (prints `[events]`), passes after.

`make check` green (lint, tsc 0 errors, 1306 tests pass, CLAUDE.md table matches). Folds into the existing `init6-events-logger-injection` changeset — no new changeset.

The WAL side of this PR was reviewed for the same class of mistake and is correct: replay forwards the raw `options?.logger` (undefined in the default path) to the WAL ops, so each op resolves its own `wal` component as before. Component names and scopes (`engram:wal`, `engram:wal-replay`, `centient:events`, `centient:events:replay`) all match the originals.

owenjohnson:
Addressed all 5 findings from the latest maintainer review (pushed `5e3e942`). `make check` green (lint + 1308 tests + claudemd-check).

**[MEDIUM] error-handling — `events/src/jsonl.ts` flush()**
Root-cause fix, not band-aid. `doFlush()` already requeued failed lines and logged, but the *public* `flush()` swallowed the rejection and resolved successfully, so a caller awaiting it for durability got a false success. `doFlush()` now **re-throws** after requeuing, and the public `flush()` propagates it (the chain is advanced through a non-throwing tail so one failure doesn't poison subsequent flushes). Fire-and-forget paths (periodic timer, eager batch, onClose, stream close/dispose) keep their existing `.catch()`-and-log and never crash the stream. Factory JSDoc documents that an awaited `flush()` now rejects on write failure.

**[LOW] test-coverage — `events/tests/logger-injection.test.ts` console.error isolation**
Replaced the manual `console.error` save/try-finally (which only restored on the happy path, before the assertions) with `vi.spyOn(console, "error").mockImplementation(...)` + `afterEach(vi.restoreAllMocks)`, so a mid-body throw can no longer leak a patched `console.error` into later tests.

**[LOW] api-contracts — `events/src/logging.ts` EventsLogger overloads**
Added an explicit "Implementer contract" section to the `EventsLogger` JSDoc: the events internals call BOTH the `(context, message)` and `(message)` forms, so a custom logger must handle both call shapes (it can't assume arg 1 is always the context). Included the idiomatic branch-on-first-arg pattern.

**[LOW] error-handling — `wal/src/wal.ts` cleanupOrphanedTempFiles**
Changed the return type from `void` to a new `WALCleanupResult` (`{ success, removed, failures }`). Per-file delete failures are still logged but are now also collected in `failures` and flip `success` to `false`, so a caller can detect e.g. a permissions problem instead of silently accumulating stale temp files. Stays best-effort/non-throwing; a missing dir still returns success. New `WALCleanupResult`/`WALCleanupFailure` types exported; README and tests updated, including a real EACCES failure-path test (read-only parent dir; skipped on win32/root).

**[LOW] test-coverage — `events/src/stream.ts` jsonl() logger forwarding**
Added a test that injects a capture logger into `createEventStream`, calls `stream.jsonl(...)`, emits a non-serializable value, and asserts the injected logger receives the `JSONL serialization failed` error — verifying the forwarding wire end-to-end, not just that the component name survives.

The two existing `init6-*` minor changesets were amended to describe the new flush-rejection and cleanup-result behaviour (both are additive within these initiatives, so they stay `minor`).

owenjohnson:
Integrated `origin/main` into this branch to clear the wave conflicts. Resolved KEEP-BOTH:

- **CLAUDE.md** (package table): the `@centient/sdk` and `@centient/wal` rows both conflicted. Kept both sides' additions — main's `events.subscribeIter()` AsyncIterable / gated `events.subscribe()` note on the `@centient/sdk` row, and this PR's `WalLogger` structural-logger-injection note on the `@centient/wal` row. No versions changed; `claudemd-check` passes.

No other files conflicted. `make check` is green (lint + tsc 0 errors + vitest 1328 passed / 14 skipped + claudemd-check). Merge commit pushed; ready for the harness to merge.

owenjohnson:
Coordinator merge-bot: merging under operator authorization. Verified: maintainer APPROVED on current head `4aac7ca` (review commit_id == head SHA), mergeable, base=main.

PR #110, opened 2026-06-15
feat(sdk): adopt @centient/resilience backoff + export isRetryableError

owenjohnson:
## What

Makes `@centient/sdk` the first day-one consumer of `@centient/resilience`: the client's retry backoff is now delegated to the shared `createBackoff` primitive, and the SDK's retryable-error decision is extracted into a new **public, exported `isRetryableError(err)` helper** that the client also uses internally.

From the public-packages plan (§1 "Day-one consumers") + the #91 follow-ups: stand up the shared backoff/retry primitives, then have the SDK consume them and expose `isRetryableError` so centient/mbot/test-kit can stop hand-rolling the classification by string-matching error messages.

## Backoff drop-in (behaviour-preserving)

The client's private `backoffDelay(attempt)` previously inlined:

```
this.retryDelay * attempt + Math.random() * this.retryDelay * 0.5
```

It now delegates to `@centient/resilience`'s `createBackoff({ baseDelayMs: retryDelay, strategy: "linear", jitterRatio: 0.5 })`, whose `delayFor(attempt)` computes:

```
attempt * baseDelayMs + random() * (jitterRatio * baseDelayMs)
  = attempt * retryDelay + random() * (0.5 * retryDelay)
```

**Algebraically identical** to the historical formula. The resilience default randomness source (`systemRandom`) calls `Math.random()`, so the existing tests that stub `Math.random` continue to pin jitter deterministically. The pre-existing jitter/backoff tests (bounds, stubbed-random midpoint = `12.5`, real-random range) pass **unmodified**; only the source-grep test that asserted the *literal old formula* was updated to assert the new delegation (no `retryDelay *` left in `client.ts`; `backoffDelay` routes through `this.backoff.delayFor(attempt)`).

## New export — `isRetryableError(err): boolean`

Encodes the client's own retry policy as one authoritative predicate (`src/retry.ts`), re-exported from the barrel:

- **Retryable** (`true`): a 5xx `EngramError`, or a raw transport `Error` (e.g. a `fetch` `TypeError` / `ECONNREFUSED`) that the client retries before wrapping into a `NetworkError` on exhaustion.
- **Non-retryable** (`false`): `TimeoutError`, `NetworkError` (deterministic non-JSON/parse failures), `ResponseShapeError`, any 4xx `EngramError`, and non-`Error` throwables.

The client's `request()` 5xx retry branch now calls `isRetryableError(error)` — **single source of truth**, no drift between the helper and the loop.

## Intra-monorepo dependency rationale

Adds `@centient/resilience` as the SDK's first `@centient -> @centient` dependency, using the repo's existing `workspace:*` convention (same as `events` and `wal` depending on `logger`). This is an **internal composition** — resilience exists precisely so the SDK consumes its backoff math — and adds **zero external runtime deps**. The `pnpm-lock.yaml` change is the single `link:../resilience` entry. Public surface is otherwise unchanged; the `isRetryableError` export is purely additive (minor changeset).

## Test evidence

- `make check` green: lint (tsc, 0 errors) + full test (vitest **1801 passed, 14 skipped**) + python-test (**546 passed**) + claudemd-check (all versions + resource count OK) + lockfile committed.
- New focused `tests/retry.test.ts` covers the classification: 5xx + raw-transport retry; timeout / 4xx / `NetworkError` / `ResponseShapeError` / non-`Error` no-retry.
- Existing SDK suite (552 tests in `packages/sdk`) passes, including the unchanged jitter/backoff tests.

### DIFF
--- a/.changeset/sdk-adopt-resilience.md
+++ b/.changeset/sdk-adopt-resilience.md
@@ -1,0 +1,50 @@
+---
+"@centient/sdk": minor
+---
+
+retry: adopt `@centient/resilience` backoff + export `isRetryableError` (behaviour-preserving)
+
+## Added — `isRetryableError(err): boolean`
+
+The SDK now exports the same predicate its request loop uses to decide whether a
+caught error is worth re-issuing, so downstream consumers (centient, mbot,
+test-kit) no longer have to hand-roll the classification by string-matching
+error messages:
+
+```typescript
+import { isRetryableError } from "@centient/sdk";
+
+try {
+  await client.search(sessionId, { query });
+} catch (err) {
+  if (isRetryableError(err)) {
+    // transient — back off and try again
+  } else {
+    throw err; // terminal
+  }
+}
+```
+
+- **Retryable** (`true`): a 5xx `EngramError`, or a raw transport `Error` (e.g. a
+  `fetch` `TypeError` / `ECONNREFUSED`).
+- **Non-retryable** (`false`): `TimeoutError`, `NetworkError`,
+  `ResponseShapeError`, any 4xx `EngramError`, and non-`Error` throwables.
+
+The client uses this helper internally for its 5xx retry decision, so there is a
+single source of truth.
+
+## Changed — backoff now powered by `@centient/resilience` (behaviour-preserving)
+
+The client's private retry backoff is now delegated to
+`@centient/resilience`'s `createBackoff` (linear strategy, `jitterRatio` 0.5)
+instead of an inlined formula. This is **behaviour-preserving**: the resilience
+linear schedule is `attempt * retryDelay + random() * (0.5 * retryDelay)` —
+identical to the schedule the client has always used — and its default
+randomness source calls `Math.random()`, so timing and jitter are unchanged and
+existing jitter tests pass unmodified.
+
+This adds `@centient/resilience` as the SDK's first intra-monorepo `@centient`
+dependency (`workspace:*`). It is an internal composition — resilience exists
+precisely so the SDK consumes it — and does not add any **external** runtime
+dependency. The public surface is otherwise unchanged (the `isRetryableError`
+export is purely additive).
--- a/packages/sdk/README.md
+++ b/packages/sdk/README.md
@@ -94,6 +94,39 @@ is genuinely down. Any retry, backoff, circuit-breaking, or fallback behavior
 calls in your own error handling and decide, per call site, whether to retry,
 surface the failure to the user, or fail the operation.
 
+The retry **backoff** is powered by [`@centient/resilience`](../resilience)'s
+`createBackoff` primitive (linear strategy, 0.5 jitter ratio) — the sleep before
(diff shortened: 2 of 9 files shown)


owenjohnson:
Addressed all 4 findings from the latest maintainer review (round_refs: 54a29a30). Comments + tests only — no behavioral change, so the existing `.changeset/sdk-adopt-resilience.md` stands.

**[MEDIUM] api-contracts — `isRetryableError` broad generic-Error classification** (`retry.ts`)
Documented the deliberately-broad branch rather than narrowing it. `fetch` (WHATWG) surfaces *all* transport failures — DNS, connection-refused, TLS, resets — as a bare `TypeError` ("Failed to fetch" / "fetch failed"). There is no structured signal separating a genuine network `TypeError` from a bug-thrown one, so excluding `TypeError`/`ReferenceError`/`SyntaxError` by constructor would silently break retries for real outages (the common transient case) to guard against the rare bug case — the wrong trade and a silent-degradation hazard (repo P-principles). Added a prominent JSDoc caveat plus an inline note at the branch, and told downstream direct callers to pre-classify those error types at their call site if it matters for them.

**[LOW] test-coverage — `createClientBackoff` lacks direct tests** (`retry.ts`)
Added 3 direct unit tests in `retry.test.ts` (12 total now, was 9): random=0 reproduces the exact historical linear base (`attempt * retryDelay`); a near-1 random pins the jitter into `[0, 0.5*retryDelay)` confirming the 0.5 jitter ratio; and a third asserts base+jitter scale with `retryDelayMs`. These validate the behaviour-preservation claim directly rather than only via `client.test.ts`.

**[LOW] error-handling — imprecise "short-circuited above" comment** (`client.ts`)
Rewrote the comment to be precise: `NetworkError` cannot reach the `isRetryableError` branch because the dedicated `error instanceof NetworkError` guard a few lines up re-throws it first, and (independently) it carries no >=500 statusCode so it would classify non-retryable anyway.

**[LOW] type-safety — `Backoff` type coupling on private property** (`client.ts`)
Smallest honest change: documented why the coupling is intentionally minimal rather than restructuring. `Backoff` is a **type-only** import (erased at build, zero runtime coupling), the field is `private` and not on the SDK's public `.d.ts` surface (no `Backoff` re-export from the barrel — verified), and the only contract consumed is `delayFor(attempt)`. Introducing a parallel local interface would duplicate the resilience contract for no runtime benefit.

`make check` green (lint + 1804 tests passed / 14 skipped + claudemd-check OK).

owenjohnson:
R2: narrowed isRetryableError to exclude programming-error constructors (ReferenceError/SyntaxError/RangeError/EvalError) per the finding, while keeping TypeError retryable since fetch surfaces transport failures as TypeError — addresses the finding in code.

owenjohnson:
cl merge-drain (break-glass, operator-authorized): merging — maintainer APPROVED on current head `8c06ade` (review commit_id == head SHA), mergeable, base=main.

PR #118, opened 2026-06-19
fix(sdk): parseApiError nested-envelope ignores statusCode switch (#117)

### LINKED ISSUE #117: parseApiError: nested-error envelope ignores statusCode switch — 404 throws base EngramError, not NotFoundError
## Summary

`parseApiError` (`src/errors.ts` / `dist/errors.js`) handles the **nested** error envelope `{ error: { code, message, details? } }` in a branch that runs **before** the `switch (statusCode)` block. For any code that isn't `SHIMMER_CAS_CONFLICT` / `SHIMMER_DISABLED`, that branch throws a **base `EngramError`** and returns — so the `case 404 → NotFoundError` (and 401/409/500) mappings **never apply to nested-envelope responses**. They only fire for the flat `{ code, message }` body shape.

## Impact (real, hit in production)

engram's shimmer GET returns "no live shimmer" as a nested envelope: `{"error":{"code":"RES_NOT_FOUND","message":"No live shimmer for this (recordType, key)"}}` with HTTP 404. The SDK throws this as a **base `EngramError`** (statusCode 404), **not** `NotFoundError`.

Consumers that catch `NotFoundError` to detect a healthy "route live, record absent" 404 therefore misclassify it. In mbot this made the ADR-031 shimmer-heartbeat migration inert: the availability probe fail-closed on every healthy 404, no shimmer heartbeats were written/read, and a fleet-wide false-`unobserved` container-restart storm never cleared. mbot worked around it by classifying on `statusCode === 404` instead of the error class, but the SDK contract is surprising: **the thrown class depends on the response envelope shape, not the HTTP status.**

## Repro

```ts
import { parseApiError, NotFoundError } from "@centient/sdk";
try {
  parseApiError(404, { error: { code: "RES_NOT_FOUND", message: "No live shimmer for this (recordType, key)" } });
} catch (e) {
  console.log(e instanceof NotFoundError); // false  ← expected true
  console.log(e.constructor.name);         // "EngramError"
  console.log(e.statusCode);               // 404
}
// Flat shape works as expected:
try {
  parseApiError(404, { code: "X", message: "y" });
} catch (e) { console.log(e instanceof NotFoundError); } // true
```

## Suggested fix

In the nested-error branch, route by `statusCode` before the generic `throw new EngramError(...)` — i.e. apply the same `case 404 → NotFoundError` / 401 / 409 / 500 mapping the flat-shape branch uses (preserving the `code`/`details`). Equivalently, factor the status→class mapping into one helper both branches call. This makes the thrown class a function of HTTP status regardless of envelope shape.

## Found via

mbot 0.28.x ADR-031 D1 shimmer heartbeat migration against engram 0.42.0 + @centient/sdk 2.1.0.

owenjohnson:
## Root cause

`parseApiError` (`packages/sdk/src/errors.ts`) handled the **nested** error
envelope `{ error: { code, message, details? } }` in a branch that ran **before**
the `switch (statusCode)` block. For any code other than `SHIMMER_CAS_CONFLICT` /
`SHIMMER_DISABLED`, that branch threw a base `EngramError` and returned — so the
`404 → NotFoundError`, `401 → UnauthorizedError`,
`409 → CrystalVersionConflictError`, `500 → InternalError` mappings **never
applied to nested-envelope responses**; they only fired for the flat
`{ code, message }` body shape.

The thrown error CLASS therefore depended on the **response envelope shape**, not
the HTTP status. In production, engram's nested `404 RES_NOT_FOUND` ("no live
shimmer") was thrown as a base `EngramError`, so consumers catching
`NotFoundError` to detect a healthy "route live, record absent" 404
misclassified it — breaking mbot's ADR-031 shimmer-heartbeat probe (found against
engram 0.42.0 + @centient/sdk 2.1.0).

## Fix

Extract one shared `errorForCode(code, message, statusCode, details, rawBody)`
helper that maps `code` + `statusCode` to the typed `EngramError` subclass, and
route **both** envelope branches (nested + flat) through it. The thrown class is
now a function of HTTP status regardless of envelope shape.

- Shimmer special-cases (`SHIMMER_CAS_CONFLICT`, `SHIMMER_DISABLED`) and the
  `OPERATION_VERSION_CONFLICT` CAS case still map to their typed errors (the
  helper checks them before the status switch).
- The server's original `code`/`details` are **preserved** on the status-keyed
  classes: `NotFoundError`/`UnauthorizedError`/`InternalError` gained optional
  `code`/`details` params (defaulting to their legacy hardcoded codes). So a
  nested `RES_NOT_FOUND` 404 is both `instanceof NotFoundError` **and** keeps
  `code === "RES_NOT_FOUND"` — satisfying the existing shimmer/notes-dedup/sync
  resource tests that assert the server code survives.
- A generic 409 (e.g. `SYNC_SCHEMA_VERSION_MISMATCH`) stays a base `EngramError`
  carrying its real code/message/details, rather than being rewritten into a
  `SessionExistsError` (whose constructor mangles the message into
  "Session … already exists"). Only the typed `SESSION_EXISTS` 409 maps to
  `SessionExistsError`.

## Tests

New `parseApiError` nested-envelope group in `packages/sdk/tests/errors.test.ts`:
nested 404 → `NotFoundError` (and `code` preserved as `RES_NOT_FOUND`),
nested 401 → `UnauthorizedError`, nested 409 `OPERATION_VERSION_CONFLICT` →
`CrystalVersionConflictError` (with `currentVersion`), nested generic 409
`SESSION_EXISTS` → `SessionExistsError`, nested 500 → `InternalError`, nested
unmapped status → `EngramError`, and both shimmer cases still mapping from the
nested shape.

## Verification

- `npm run build` (tsc): clean
- `npm test`: **616 passed (27 files)**
- `npm run lint` (tsc --noEmit): clean

Patch-bump changeset added: `.changeset/fix-parseapierror-nested-envelope.md`.

Closes #117

### DIFF
--- a/.changeset/fix-parseapierror-nested-envelope.md
+++ b/.changeset/fix-parseapierror-nested-envelope.md
@@ -1,0 +1,29 @@
+---
+"@centient/sdk": patch
+---
+
+Fix `parseApiError` so the thrown error CLASS is a function of the HTTP status,
+not the response envelope shape (issue #117).
+
+The nested error envelope `{ error: { code, message, details? } }` was handled
+in a branch that ran before the `switch (statusCode)` mapping and — for every
+code other than the two shimmer special-cases — threw a base `EngramError`. As
+a result the `404 → NotFoundError` / `401 → UnauthorizedError` /
+`409 → CrystalVersionConflictError` / `500 → InternalError` mappings only fired
+for the flat `{ code, message }` body shape, never for the nested envelope. In
+practice engram's nested `404 RES_NOT_FOUND` ("no live shimmer") was thrown as a
+base `EngramError`, so consumers catching `NotFoundError` to detect a healthy
+"route live, record absent" 404 misclassified it (broke mbot's ADR-031
+shimmer-heartbeat probe against engram 0.42.0 + @centient/sdk 2.1.0).
+
+Both envelope branches now route through one shared `errorForCode` helper that
+maps `code` + `statusCode` to the typed class, so a nested-envelope 404 is
+`instanceof NotFoundError` exactly like a flat-body 404. The typed shimmer
+special-cases (`SHIMMER_CAS_CONFLICT`, `SHIMMER_DISABLED`) and the
+`OPERATION_VERSION_CONFLICT` CAS case still map to their typed errors. The
+server's original `code` and `details` are preserved on the status-keyed classes
+(`NotFoundError`/`UnauthorizedError`/`InternalError` gained optional
+`code`/`details` params), so a nested `RES_NOT_FOUND` 404 is both
+`instanceof NotFoundError` AND keeps `code === "RES_NOT_FOUND"`. A generic 409
+(e.g. `SYNC_SCHEMA_VERSION_MISMATCH`) stays a base `EngramError` carrying its
+real code/message/details rather than being rewritten into a `SessionExistsError`.
--- a/.changeset/fix-parseapierror-nested-envelope.md
+++ b/.changeset/fix-parseapierror-nested-envelope.md
@@ -1,5 +1,5 @@
 ---
-"@centient/sdk": patch
+"@centient/sdk": minor
 ---
 
 Fix `parseApiError` so the thrown error CLASS is a function of the HTTP status,
@@ -27,3 +27,14 @@ server's original `code` and `details` are preserved on the status-keyed classes
 `instanceof NotFoundError` AND keeps `code === "RES_NOT_FOUND"`. A generic 409
 (e.g. `SYNC_SCHEMA_VERSION_MISMATCH`) stays a base `EngramError` carrying its
 real code/message/details rather than being rewritten into a `SessionExistsError`.
+
+**Minor (not patch) — additive public-API surface.** The public error
+constructors `NotFoundError`, `UnauthorizedError`, and `InternalError` gain two
+optional trailing params, `(message, code?, details?)`, so the server's original
+`code`/`details` can be preserved when routed through `parseApiError`. The change
+is backward-compatible — the `code` defaults match the previously-hardcoded
+values (`NOT_FOUND`, `UNAUTHORIZED`, `INTERNAL_ERROR`) and `details` defaults to
+`undefined` — so existing call sites and `new NotFoundError(msg)` usages behave
+identically. Because the public type signatures of exported error classes
+changed (a new, larger callable surface), this is released as a **minor** bump
+rather than a patch.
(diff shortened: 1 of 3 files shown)


owenjohnson:
Addressed all 4 maintainer-review findings in commit 8513975:

**1. [MEDIUM] api-contracts — changeset bump.** Changed `.changeset/fix-parseapierror-nested-envelope.md` from `patch` → `minor` for `@centient/sdk` and added a note documenting the additive surface: `NotFoundError`, `UnauthorizedError`, and `InternalError` gained optional trailing `(code?, details?)` params. Verified backward-compatibility — the `code` defaults match the previously-hardcoded values (`NOT_FOUND` / `UNAUTHORIZED` / `INTERNAL_ERROR`) and `details` defaults to `undefined`, so `new NotFoundError(msg)` and existing call sites behave identically.

**2. [LOW] error-handling — defensive body read.** `errorForCode` now reads `currentVersion` off the raw 409 body through a new non-throwing `readNumericProp` helper (try/catch around the property access). A null-prototype object or a throwing proxy can no longer make `parseApiError` itself throw — it falls back to `NaN` (the existing "version unknown" signal).

**3. [LOW] test-coverage — added tests** in `packages/sdk/tests/errors.test.ts`: (a) nested envelope with a `details` field — asserts `details` is preserved onto the thrown `NotFoundError`; (b) `OPERATION_VERSION_CONFLICT` with a non-numeric `currentVersion` — asserts it drops to `NaN`, not a bogus version (plus a null-prototype-body guard test); (c) a generic non-`SESSION_EXISTS` 409 (`SYNC_SCHEMA_VERSION_MISMATCH`) — asserts it stays a base `EngramError` carrying its real code/message, not a `SessionExistsError`.

**4. [LOW] error-handling — comment accuracy.** Verified the actual behavior: `SessionExistsError`'s constructor takes a single `sessionId` arg and wraps it into the fixed template `Session <arg> already exists` — it does *rewrite*, but it does NOT preserve an arbitrary server message. Reworded the inline comment to state this accurately and to explain *why* only `SESSION_EXISTS` routes to it (any other 409 would get its message mangled and `details` dropped).

Verification: `pnpm build` ✅, `npm test` ✅ 620 passed (616 prior + 4 new), `pnpm lint` (`tsc --noEmit`) ✅ clean. No version bumped by hand; pushed to the existing branch only.

owenjohnson:
cl merge-drain (break-glass, operator-authorized): merging — maintainer APPROVED on current head `8513975` (review commit_id == head SHA), mergeable, base=main.

PR #114, opened 2026-06-16
feat(sdk): public dedup/merge resource methods (#81)

owenjohnson:
Closes #81.

Exposes the dedup/merge-review endpoints previously reachable only through the `@internal` `_request` helper as public, typed resource methods on the existing resource classes. This is the SDK side of #81's retirement path; the centient migration off `asInternal()` is a separate centient-repo PR.

## Endpoints covered

| Endpoint | Method | Notes |
|----------|--------|-------|
| `POST /v1/notes/:id/dedup` | `notes.dedup(id, { mergeMethod?, threshold? })` | Bare (non-enveloped) `{action, mergeId, confidence, canonicalId}`; snake_case wire fields normalized; body mapped to the server's `.strict()` snake_case schema |
| `GET /v1/crystals/merges/pending` | `crystals.pendingMerges({ sessionId?, limit? })` | Bare `{success, pending, total}` |
| `POST /v1/crystals/merges/:id/review` | `crystals.reviewMerge(mergeId, { decision, mergedContent? })` | Bare `{success, ...result}`; `mergedContent` → snake_case `merged_content` |
| `GET /v1/crystals/merges/history/:id` | `crystals.mergeHistory(itemId)` | Bare `{success, id, merge_chain, total}` |

All methods route through the same `request()` seam existing methods use (no re-exposed `_request`), with `#106`'s `validate.ts` guards (`requireObject`/`requireField`/`requireArray`) applied at the HTTP edge since the merge routes return **bare** `{ success, ... }` objects rather than the standard `{ data }` envelope. Wire shapes were taken from the authoritative engram-server route handlers (`session-notes.ts`, `crystals.ts`) and service types (`PendingMerge`, `ReviewMergeResult`, `MergeRecord`).

## Flagged — not implemented (no server route exists)

The remaining endpoints named in #81 have **no route handler in engram-server**, so no SDK method was added (implementing against a non-existent route would be guessing):

- **Consolidation lifecycle** (`POST/GET/PATCH consolidation-events`, `GET consolidation-queue`, `DELETE undo`) — the repository/service/table exist (`repositories/consolidation-events.ts`, `services/consolidation.ts`) but no `/v1/sessions/:id/consolidation-*` HTTP routes are registered.
[...]

PR #112, opened 2026-06-15
feat(sdk): add Shimmers client for engram /v1/shimmers (closes #109)

owenjohnson:
Closes #109.

Adds `client.shimmers` to `@centient/sdk`, wrapping engram's now-live `/v1/shimmers` API (ADR-027; engram-server #931 / #933). A *shimmer* is node-local, TTL-backed operational state — what engram is *doing right now* — with three record types: `lock` (CAS), `heartbeat` (overwrite), `ipc` (write-once + exactly-once consume).

### Methods → endpoint mapping

| Method | Endpoint | Semantic |
|---|---|---|
| `heartbeat(key, value, { ttlSeconds })` | `POST /v1/shimmers` (heartbeat) | unconditional overwrite + TTL (never CAS) |
| `acquireLock(key, { ownerToken, ttlSeconds, value? })` | `POST /v1/shimmers` (lock) | acquire-if-free CAS; echoes caller's `ownerToken` |
| `renewLock(key, { ownerToken, expectedRevision, ttlSeconds, value? })` | `PUT /v1/shimmers/:key` (lock) | renew/CAS a held lock |
| `releaseLock(key, { ownerToken })` | `DELETE /v1/shimmers/:key?recordType=lock&ownerToken=…` | owner-guarded release |
| `emitIpc(key, value, { ttlSeconds })` | `POST /v1/shimmers` (ipc) | write-once post |
| `consumeIpc(key)` | `DELETE /v1/shimmers/:key?recordType=ipc` | atomic exactly-once; returns the record or `null` |
| `get(key, recordType)` | `GET /v1/shimmers/:key?recordType=…` | TTL-filtered read |

### Error mapping (errors.ts)

- **409 `SHIMMER_CAS_CONFLICT` → `ShimmerCasConflictError`** — lock contention (acquire/renew) or ipc write-once collision. The holder's `ownerToken` is never exposed (the 409 body is the redacted projection).
- **503 `SHIMMER_DISABLED` → `ShimmerDisabledError`** — the surface is gated behind `ENGRAM_SHIMMER_ENABLED` (default off); not retried.

Both new codes added to the `ErrorCode` union and routed in `parseApiError` (nested Hono envelope + bare `{ code, message }`).

### Contract notes / ambiguities resolved against the engram route source

[...]

PR #119, opened 2026-06-19
feat(sdk-python): Phase C — full @centient/sdk 2.1.0 parity + nested-envelope error fix (#117 sibling)

owenjohnson:
## Summary

Completes **sdk-python Phase C**: brings `engram-py` to full `@centient/sdk` 2.1.0 **resource parity** and aligns the version `1.0.0 → 2.1.0`. Also fixes the Python sibling of issue #117 (nested-envelope error mapping), folded in per request.

### Phase C — new resources (async + synchronous-client variants, wired into both clients)

- **`sync`** (`client.sync`): NDJSON `push`/`pull`, `push_to`/`pull_from`, `get_status`, `list_conflicts`/`resolve_conflict`, and a `.peers` sub-resource. Standard `{data}` envelopes + bare `{peer}` peer shapes; malformed NDJSON → `NetworkError`, contract drift → `EngramError`.
- **`agents`, `ambient_context`, `facts`, `gc`, `memory_spaces`, `users`** — the six plan resources.
- **`shimmers`** (`client.shimmers`): lock / heartbeat / ipc, with typed `ShimmerCasConflictError` / `ShimmerDisabledError` (post-plan surface, #112).
- **Crystal dedup / deferred-merge review** (post-plan, #114): `notes.dedup`, `crystals.pending_merges` / `review_merge` / `merge_history` — bare (non-enveloped) bodies.

### Supporting changes

- `EngramError.retryable` property (default `True`; `False` on `ShimmerDisabledError`) now honored across the client retry loop, so the 503 deployment gate is not retried (mirrors TS).
- Version `1.0.0 → 2.1.0`; CHANGELOG support matrix + README updated to 2.1.0 (every TS resource surface now has a Python counterpart; flat client methods remain out of scope by design).
- `types/__init__` imports `sync` before `terrafirma` so terrafirma keeps the flat-namespace `SyncStatus`; the replication one stays at `engram.types.sync.SyncStatus`.

### #117 sibling fix (`parse_api_error`)

The nested `{ error: { code, message } }` branch threw a base `EngramError` for every non-CAS/non-shimmer code, never reaching the status→class mapping (404→`NotFoundError`, etc.) — exactly the TS bug fixed in #118. Extracted a shared `_status_mapped_error` helper routed through by **both** envelope branches; server code/details preserved; generic 409 (non-`SESSION_EXISTS`) now stays a base `EngramError` in both branches.

## Testing

`make python-test` (`scripts/run-python-tests.sh`): **668 passed, 10 deselected**. ~120 new tests (every new resource + `TestNestedEnvelopeStatusMapping`). The 10 deselected integration tests (autogen/crewai/langchain) are pre-existing and excluded from the gate.

## Notes

- `engram-py` is a PyPI package, not in the npm Changesets flow — version bumped in `pyproject.toml` per the Python release procedure (no changeset).
- Relates to TS #117 / #118 (the TS-side fix of the same error-mapping bug).

### DIFF
--- a/packages/sdk-python/CHANGELOG.md
+++ b/packages/sdk-python/CHANGELOG.md
@@ -1,34 +1,60 @@
 # Changelog
 
-## Unreleased
+## 2.1.0 (Phase B + C — full `@centient/sdk` 2.1.0 resource parity)
 
-### Parity correction — TypeScript SDK 2.0.0
+This release closes the parity gap with the TypeScript SDK: the version is now
+aligned to `@centient/sdk` 2.1.0, and every TS resource surface has a Python
+counterpart (or an explicit line in the support matrix saying why not).
 
-> **Correction.** Earlier releases of this CHANGELOG (the 1.0.0 entry below)
-> claimed *"Full API parity"* with the TypeScript SDK. That claim was true at
-> the time but went stale when `@centient/sdk` moved 1.7.1 → 2.0.0 underneath
-> this package. **There is no full-API-parity guarantee.** Parity is now tracked
-> explicitly in the support matrix below and in `README.md`. This package does
-> not claim parity it does not have.
-
-#### Support matrix (vs `@centient/sdk` 2.0.0)
+#### Support matrix (vs `@centient/sdk` 2.1.0)
 
 | TS SDK surface | Python? | Notes |
 |----------------|---------|-------|
 | Sessions, notes, scratch, coordination | yes | `client.sessions` and sub-resources |
 | Crystals (CRUD, search, ACL, share, fork, hierarchy, versions, trash, merge, clusters) | yes | `client.crystals` |
-| `expected_version` CAS on `crystals.update` | yes | **new** — mirrors TS `expectedVersion`; 409 maps to `CrystalVersionConflictError` |
-| `skip_embedding` on `crystals.create` / `crystals.update` | yes | **new** — mirrors TS `skipEmbedding`; server-floor caveats documented |
-| `MaintenanceResource` (`vacuum`, `tombstone_cleanup`, `changelog_compact`) | yes | **new** — handles bare (non-enveloped) bodies |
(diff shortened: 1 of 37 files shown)


owenjohnson:
Addressed the maintainer review in 46f4705:

- **[MEDIUM] retry uses `exc.retryable` without a guard** — now `getattr(exc, "retryable", True)` at all retry sites. (The retry blocks only catch `EngramError`, which always defines `retryable`, but the guard makes it robust against any future exception type reaching that path.)
- **[LOW] NDJSON wrong-typed field** — `_parse_pull_ndjson` now wraps `SyncChange.model_validate` and re-raises a structured `NetworkError` with the line index, so a type drift surfaces consistently with the other parse-boundary failures instead of leaking a raw pydantic error.
- **[LOW] `SyncStatus` import-order fragility** — removed the order dependency: the flat `engram.types.SyncStatus` is now bound to terrafirma's explicitly via `from engram.types.terrafirma import SyncStatus as SyncStatus` at the end of `types/__init__`. Resolution no longer depends on star-import order; the replication type stays at `engram.types.sync.SyncStatus`.
- **[LOW] NDJSON malformed-line test** — strengthened to assert the line index + route appear in the message, and added a wrong-field-type case.

Two findings I kept as-is by design (no code change):

- **`_require_bare` checks for a `data` key to detect envelope drift** — this is the established `maintenance.py` idiom for the bare merge-review routes (their bodies are `{success, pending, total}` etc.; a legitimate bare body never carries a top-level `data`). Diverging here would split the bare-body convention across two patterns.
- **NDJSON error truncates the line at 200 chars** — bounded error messages are intentional (matches the `maintenance.py` `_truncate_body` cap) so a large/sensitive payload can't be dumped into a log line; the line index + route already pinpoint the failure.

Gate: `make python-test` → 669 passed, 10 deselected.

owenjohnson:
cl merge-drain (break-glass, operator-authorized): merging — maintainer APPROVED on current head `4343fc0` (review commit_id == head SHA), mergeable, base=main.

Issue #121, opened 2026-06-23
ADR-002: libsecret D-Bus migration silently reverts to insecure secret-tool path (P0)


owenjohnson:
**Source:** scoped \`cl-adr-audit\` re-run (2026-06-23). "built" claim demoted to functionally-unverified under refutation.

## Gap
0.5.0 Item 4 migrated libsecret enumeration from \`secret-tool\` (which emits decrypted values on stdout) to a D-Bus path. But \`listKeysViaDbus\` via dynamic \`import('dbus-next')\` (\`packages/secrets/src/vault-libsecret.ts:159-160\`) is **code-present only**, and the bare \`catch { return listKeysViaSecretTool() }\` at \`vault-libsecret.ts:148-150\` swallows **all** D-Bus errors and silently reverts to the secret-tool path the migration was meant to eliminate. The D-Bus path uses an \`as any\` proxy (line 185); no runtime/CI proof it ever executes.

## Impact
The security fix may be **dead-on-arrival in prod with zero signal** — if the D-Bus path throws for any reason, it silently no-ops back to the insecure path.

## Fix
Either (a) prove the D-Bus path executes (integration test or CI with \`dbus-next\` + a session bus) and replace the catch-all with error-classified handling that does **not** silently fall back on a real D-Bus failure, or (b) emit a loud one-time warning when reverting to the secret-tool path.

_Filed from session 2026-06-22-post-wave-followups (handoff Priority 4 / ADR-002 pillars)._

owenjohnson:
Fixed in PR #124 (merged to main as commit on `main`; mbot ✅ Ready). The before-hook rejection auditing (#120) and libsecret D-Bus degradation warning (#121) shipped together. Released via the `@centient/secrets` minor changeset.

Issue #120, opened 2026-06-23
ADR-002: before-hook rejection leaves credential op silently un-audited (P0)


owenjohnson:
**Source:** scoped \`cl-adr-audit\` re-run (2026-06-23). Survives adversarial refutation, high confidence.

## Gap
ADR-002 (line 158) commits: if a before-hook throws, the op aborts **AND** "after hooks on already-entered policies still run with a failure event — ensuring rejected operations are still audited."

**Code does not do this.** \`runBeforeHooks\` (\`packages/secrets/src/policy.ts:84-88\`) and all four call sites (\`vault.ts:141,162,196,234\`) have **no try/catch** — a throwing before-hook propagates directly with zero after-hook / failure event fired. The \`policy.ts\` header docstring even contradicts the ADR ("If any throws, the operation is aborted and the error propagates to the caller"). Test \`policy.test.ts:220\` confirms the abort but asserts **no** audit event.

## Impact
A denied/rejected credential operation leaves **no audit trace** — the exact auditability gap ADR-002 warns against. Three artifacts, three stories (ADR / docstring / test) — the divergence hid because no layer asserted the ADR's guarantee.

Note: read/enumerate *backend-throw* failures DO fire after-hooks via try/catch (\`vault.ts:164-176,236-248\`). The gap is specifically the **before-hook rejection** path.

## Fix
Wrap before-hook invocation so a rejection fires after-hooks with a failure event (per ADR line 158); reconcile the contradicting \`policy.ts\` docstring; add a test asserting a denied op emits an audit event.

_Filed from session 2026-06-22-post-wave-followups (handoff Priority 4 / ADR-002 pillars)._

owenjohnson:
Fixed in PR #124 (merged to main as commit on `main`; mbot ✅ Ready). The before-hook rejection auditing (#120) and libsecret D-Bus degradation warning (#121) shipped together. Released via the `@centient/secrets` minor changeset.

PR #124, opened 2026-06-23
fix(secrets): audit policy-rejected ops + surface libsecret D-Bus fallback (#120, #121)

owenjohnson:
Closes **#120** and **#121** — the two ADR-002 P0 gaps surfaced by the scoped `cl-adr-audit` re-run (both survived adversarial refutation).

## #120 — policy-rejected operations are now audited (ADR-002 §1.0.0)

Before: a `before` policy hook that rejected an operation by throwing propagated to the caller but fired **no `after` event** — a denied credential op left no audit trace, the exact gap ADR-002 line 158 warns against. The `policy.ts` docstring even contradicted the ADR.

After: `runBeforeHooks` implements the documented behavior — on a before-hook throw at policy `i`, the `after` hooks of the **already-entered** policies (`0..i-1`, bottom-to-top) fire with a new `*_rejected` event, then the original error re-throws. The rejecting policy's own `after` hook is skipped (its `before` did not complete). Wired through both the cascade vault (`store`/`get`/`delete`/`listCredential`) and the session vault's `withAudit` helper.

- `SecretsEventType` gains `credential_{read,write,delete,enumerate}_rejected`.
- New exported `rejectedEventType(op)` is the single source of truth for the op→event mapping.

## #121 — libsecret D-Bus→secret-tool fallback is no longer silent

The `secret-tool search` fallback briefly materializes secret values on stdout, so reverting to it from the secure D-Bus enumeration path is a security-relevant degradation. The bare `catch {}` now emits a **one-time stderr warning when a session bus is advertised (`DBUS_SESSION_BUS_ADDRESS` set) yet the D-Bus path still fails** — the path that should have worked didn't. A genuinely bus-less host (SSH/headless) stays quiet, since the fallback is the documented, expected behavior there.

## Tests

- `policy.test.ts` — rejection fires entered-only after-hooks; first-policy throw is a no-op; per-operation rejected event type; routing through the `auditTrail` built-in.
- `session-vault.test.ts` — a policy-denied write is audited via `withAudit` (`credential_write_rejected`, `backend: session-vault`).
- `vault-libsecret.test.ts` — warns once when bus advertised + D-Bus fails; stays quiet on a bus-less host. Existing fallback tests pinned bus-less for determinism.

234 secrets tests pass; `tsc --noEmit` clean. Changeset: `@centient/secrets` minor (additive event types + exported helper).

## Notes
- No ADR-002 doc change: the code now **conforms** to its line-158 commitment (the audit flagged code↛ADR divergence, not an ADR error). The contradicting `policy.ts`/`vault.ts` docstrings were reconciled.

### DIFF
--- a/.changeset/secrets-audit-rejected-ops-and-dbus-warning.md
+++ b/.changeset/secrets-audit-rejected-ops-and-dbus-warning.md
@@ -1,0 +1,36 @@
+---
+"@centient/secrets": minor
+---
+
+Close two ADR-002 P0 audit gaps surfaced by the scoped `cl-adr-audit` re-run:
+policy-rejected operations are now audited, and a security-relevant libsecret
+enumeration fallback is no longer silent.
+
+**#120 — denied operations are audited (ADR-002 §1.0.0).** Previously, when a
+`before` policy hook rejected an operation by throwing, the error propagated to
+the caller but **no `after` event fired** — a denied credential operation left
+no audit trace, the exact gap ADR-002 line 158 warns against (the `policy.ts`
+docstring even contradicted the ADR). `runBeforeHooks` now implements the
+documented behavior: on a before-hook throw at policy `i`, the `after` hooks of
+the already-entered policies (`0..i-1`, bottom-to-top) fire with a new
+`*_rejected` event, then the original error is re-thrown. The rejecting policy's
+own `after` hook does not fire (its `before` did not complete). Wired through
+both the cascade vault (`storeCredential`/`getCredential`/`deleteCredential`/
+`listCredentials`) and the session vault's `withAudit` helper.
+
+**Minor (not patch) — additive public surface.** `SecretsEventType` gains four
+members: `credential_read_rejected`, `credential_write_rejected`,
+`credential_delete_rejected`, `credential_enumerate_rejected`. A new
+`rejectedEventType(op)` helper (operation → rejected-event-type) is exported as
+the single source of truth for the mapping. Existing event types and emission
+are unchanged, so consumers that only `switch` on the prior types keep working.
+
+**#121 — libsecret D-Bus→secret-tool fallback is no longer silent.** The
+`secret-tool search` fallback briefly materializes secret values on stdout, so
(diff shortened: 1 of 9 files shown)


owenjohnson:
Addressed review (pushed `c2e2a13`):

**Codex P2 — `rejectedEventType` not exported from root:** Fixed. Re-exported from `index.ts`, so `import { rejectedEventType } from "@centient/secrets"` now resolves and the changeset's "additive public API" claim is accurate.

**mbot MEDIUM — async after-hook rejection could leak:** The `SecretsPolicy.after` signature is `(event) => void` (synchronous by contract — unlike `before`, which is `void | Promise<void>` and is awaited). The runner catches synchronous throws only; an async after-hook is not a supported shape. Per the recommendation's accepted alternative, I documented this explicitly on `SecretsPolicy.after` and `runAfterHooksForRange` (do audit I/O via a fire-and-forget sink the hook owns). No behavior change — the prior `runAfterHooks` already invoked hooks synchronously; this PR did not introduce the sync invocation.

**mbot LOW — exhaustiveness:** `rejectedEventType` now has a `never`-typed default; adding a `SecretsOperation["type"]` without a case is a compile error.

**mbot LOW — test coverage:** the rejection-event test now asserts `backend`, a non-negative `durationMs`, and a valid ISO-8601 `timestamp` (not just type/key/error).

**mbot LOW — broad libsecret catch:** Kept intentionally broad and documented why — the contract is "if the secure D-Bus enumeration fails for any reason, degrade to secret-tool rather than fail the read." The dominant cause is a bus-less host, but a protocol/variant error should also fall back. Critically, the degradation is **not silent**: `warnOnDbusDegradation` emits a one-time warning when a session bus is advertised yet D-Bus still fails, so a swallowed cause remains observable. Narrowing to specific error classes (e.g. excluding OOM) would be fragile and OOM is rarely meaningfully catchable here.

owenjohnson:
cl merge-drain (break-glass, operator-authorized): merging — maintainer APPROVED on current head `c2e2a13` (review commit_id == head SHA), mergeable, base=main.

Issue #102, opened 2026-06-13
secrets: vault key hex visible in ps via op item create/edit argv — move to 1Password Connect/SDK


owenjohnson:
**ADR-001 documented negative — tracking ticket.** Opened by Initiative 5 of `docs/plans/2026-06-12-next-stage.md`. Small, scoped.

## ADR section
- `docs/adr/001-key-provider-abstraction.md` → **Consequences → Negative → "Process argument visibility"**: "hex key is passed as a CLI argument to `op item create/edit`, visible in `ps` output momentarily. This is consistent with the existing `security` CLI approach. A future improvement could use 1Password Connect or SDK for process-internal key handling."

## Problem
`OnePasswordProvider.storeKey()` shells out to `op item create`/`op item edit` with the 32-byte vault key as a hex string in argv. For the storeKey window, any local process can read the key from `ps`/`/proc/<pid>/cmdline`. The same residue class as the `security` CLI, but the ADR flagged it as a known negative to fix.

## Design principles (by number)
- **P15 (Secure by Default)** — a secret should not transit a world-readable channel (process argv) even momentarily.
- **P16 (Authority Outside the Sandbox)** — argv exposure widens who can read the high-value vault key; the fix narrows the credential's reachable surface.

## Fix
Replace argv-passed key material with a process-internal path:
- **1Password Connect** (HTTP API to a local Connect server, key in request body over localhost), or
- **1Password SDK** (`@1password/sdk`) for in-process item create/edit with no CLI argv.

ADR-001's "No new dependencies" stance (shells out to `op` to avoid `@1password/sdk`) is the tension to resolve here: the security fix may justify the dependency. Decide and record in an ADR-001 amendment.

## Files
- `packages/secrets/src/key-providers/onepassword-provider.ts` (the `storeKey()` argv path).
- `docs/adr/001-key-provider-abstraction.md` (amendment recording the dependency tradeoff if `@1password/sdk` is adopted).
- CHANGELOG + changeset if behavior changes.

## Acceptance criteria
- Storing the vault key via the 1Password provider no longer places key hex in any process's argv (verify: no key material in `ps`/`/proc/<pid>/cmdline` during a store, asserted by test or documented manual check).
- If a dependency is added, ADR-001 amended to record the tradeoff vs its original "No new dependencies" stance.
- `make check` green.

**Inputs / related:** ADR-001 Consequences → Negative; reconciliation ticket (1Password as key- vs storage-provider).

PR #125, opened 2026-06-25
docs(adr-004): propose 1Password credential backend (OnePasswordVault)

owenjohnson:
**Status: Proposed** — design artifact for review (not an accepted decision yet).

Adds `docs/adr/004-1password-credential-backend.md`: a design for a new `OnePasswordVault implements VaultBackend`, so credential **values** (`storeCredential`/`getCredential`) can live in 1Password.

## Why

1Password is already supported — but only at the **KeyProvider** layer (ADR-001's `OnePasswordProvider` backs the vault *encryption key*). The **credential-value** layer (the Keychain→Windows→Libsecret→GPG→Env cascade) has no 1Password backend. This ADR fills that gap.

## Key decisions

- **Explicit opt-in, never auto-cascade.** `op` being installed must not silently route engram tokens into a personal vault (P15/P2). Introduces the cascade's first config seam (`secrets.backend: "1password"`), reusing `loadConfig()`.
- **Separate `secrets.onePasswordBackend` config block**, distinct from ADR-001's `secrets.onePassword` (key item) — the two layers are independent. Credential **vault is required, no default, fail-closed** when unset.
- **Secure write over stdin JSON** (`op item create -`) — secret values **never** in argv. Fixes the issue #102 leak class and gives a pattern portable back to the key provider.
- **Metadata-only enumeration** (`op item list --format=json`, no values) + **list-cache only, never cache secret values** (security + staleness reasoning in §8).
- **Shared `op` helper** factored from `OnePasswordProvider` (P6); **zero new runtime deps**; an **opt-in integration test** so the backend isn't perpetually code-present-only (the gap the ADR-001 audit flagged).

## Resolved during review with the operator
1. Config layering vs. ADR-001 → separate block.
2. Vault name → required, no default.
3. Caching → list-cache only.

One implementation-time unknown remains (whether the pinned `op` supports `op item edit -` stdin, to optimize update away from delete-then-create).

Doc-only; no code change. Becomes a `SecretsProvider` when ADR-002's 1.0 SPI lands.

### DIFF
--- a/docs/adr/004-1password-credential-backend.md
+++ b/docs/adr/004-1password-credential-backend.md
@@ -1,0 +1,344 @@
+# ADR-004: 1Password Credential Backend (`OnePasswordVault`)
+
+**Status:** Proposed
+**Date:** 2026-06-24
+**Deciders:** Owen Johnson
+**Principles:** P2 (No Silent Degradation), P3 (Transparent Evolution), P6 (Single Source of Truth), P9 (Composability Over Completeness), P15 (Secure by Default), P16 (Authority Outside the Sandbox)
+**Supersedes:** none
+**Related:** ADR-001 (Key Provider Abstraction — the *existing* 1Password support, at a different layer), ADR-002 (Long-Term Architecture — this backend becomes a `SecretsProvider` in 1.0); issue #102 (`op` argv visibility)
+
+## Context
+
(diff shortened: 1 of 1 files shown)


owenjohnson:
cl merge-drain (break-glass, operator-authorized): merging — maintainer APPROVED on current head `000462b` (review commit_id == head SHA), mergeable, base=main.

